PR #15, opened 2017-12-12
Exports props in TS definition

NejcZdovc:


brave-ui/index.d.ts
```diff
@@ -14,110 +15,113 @@ declare module "brave-ui" {
...
-          width?: string
-          height?: string
-          textColor?: string
-          background?: string
-      }
+    interface Grid {
+      id?: string
+      disabled?: boolean
+      children: React.ReactNode
+      style?: any // TODO what should be here?
```
cezaraugusto: btw this will be removed on the next minor release

PR #19, opened 2018-02-22
CI improvements

### LINKED ISSUE #17: travis improvements
the travis "passing" badge in https://github.com/brave/brave-ui just links to https://github.com/brave/brave-ui itself. it should link to https://travis-ci.org/brave/brave-ui.

travis seems to only run the tests; it should also run lint and security tests

travis also appears to be using out-of-date node and npm versions btw

cezaraugusto:
close #17
-
* badge now links to Travis
* run node on latest
* run check-security
* run lint
* run CI against master only

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,12 @@
 language: node_js
 node_js:
-  - "7"
+  - "8"
+  - "node"
 cache: npm
+script:
+  - npm run test-all
+  - npm run check-security
+  - npm run lint
+branches:
+  only:
+    - master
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,4 +1,12 @@
...
+  - "8"
+  - "node"
 cache: npm
+script:
+  - npm run test-all
+  - npm run check-security
+  - npm run lint
+branches:
+  only:
+    - master
```
diracdeltas: i think you want to run travis on all branches since travis should be required for PR

cezaraugusto: yes, we do it. The config restricts building branches but does not stop building pull-requests.

May seem overkill now as the project is too small but took advantage of this PR to make it. fyi browser-laptop shares [a similar config](https://github.com/brave/browser-laptop/blob/master/.travis.yml#L17).

cezaraugusto:
thanks!

PR #35, opened 2018-06-05
Travis and package optimizations

NejcZdovc:
- added travis parallel jobs
- remove duplicated node definitions
- replaces nsp with audit
- removed cpx as I couldn't fix it to be used
- renamed some tasks so that they are grouped together
- removed check security for push as it's expensive
- added trusted distribution
- addressed audit problems that were found
- updated package-lock
- removed babel as I don't think it's used

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -24,7 +24,7 @@ There are several ways you can do so:
 2. cd `brave-ui`
 3. `npm i` or `yarn install`
 
-Once the installation is finished, run `npm start`. The source is available under http://localhost:9091 (one after nine-o-nine?). It will hot-reload the style guide at every code change.
+Once the installation is finished, run `npm storybook-start`. The source is available under http://localhost:9091 (one after nine-o-nine?). It will hot-reload the style guide at every code change.
 
 ## Running tests
 
--- a/package.json
+++ b/package.json
@@ -12,18 +12,26 @@
   ],
   "license": "MPL-2.0",
   "scripts": {
-    "bootstrap": "lerna bootstrap",
+    "ts": "tsc",
+    "ts-watch": "tsc -w",
+    "lint": "standard --verbose",
+    "lint-fix": "standard --fix --verbose",
     "build": "lerna exec --parallel -- tsc",
-    "transpile": "tsc",
-    "dev": "tsc -w",
-    "check-security": "nsp check",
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -12,19 +12,28 @@
   ],
   "license": "MPL-2.0",
   "scripts": {
+    "ts": "tsc",
+    "ts-watch": "tsc -w",
+    "lint": "standard --verbose",
+    "lint-fix": "standard --fix --verbose",
     "bootstrap": "lerna bootstrap",
```
NejcZdovc: would love to group this 3 together as well. When do we use bootstrap and build?

NejcZdovc: @cezaraugusto

cezaraugusto: we're not using bootstrap anymore so it's more like a leftover, I'm ok grouping other tasks

docs/contributing.md
```diff
@@ -24,7 +24,7 @@ There are several ways you can do so:
 2. cd `brave-ui`
 3. `npm i` or `yarn install`
 
-Once the installation is finished, run `npm start`. The source is available under http://localhost:9091 (one after nine-o-nine?). It will hot-reload the style guide at every code change.
+Once the installation is finished, run `npm story-start`. The source is available under http://localhost:9091 (one after nine-o-nine?). It will hot-reload the style guide at every code change.
```
cezaraugusto: per script it's `storybook-start` now, right?

package.json
```diff
@@ -12,19 +12,28 @@
...
-    "publish": "lerna publish && npm run deploy",
-    "start": "start-storybook -p 9091 -c .storybook",
-    "test": "jest -t",
-    "test-all": "jest --verbose",
+    "publish": "lerna publish && npm run storybook-deploy",
+    "storybook-start": "start-storybook -p 9091 -c .storybook",
+    "storybook-deploy": "storybook-to-ghpages -s public -o .out",
+    "test-unit": "jest -t",
+    "test-unit-all": "jest --verbose",
+    "test-security": "npm audit",
```
cezaraugusto: cool stuff

docs/contributing.md
NejcZdovc: fixed

package.json
NejcZdovc: removed

PR #36, opened 2018-06-08
Rewards

NejcZdovc:
TODO

- [x] remove local states
- [x] rename onClick to more meaningful name 
- [x] add locals
- [x] optimize font family (move it to the top wrapper if used a lot)

Phase 2
- [ ] add cursor pointer to all buttons/actions
- [ ] pixel perfect
- [ ] update to the latest design
- [ ] add transitions
- [ ] proper icons


components/rewards/alert/assets/success.svg
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="42px" height="42px" viewBox="0 0 42 42" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 50.2 (55047) - http://www.bohemiancoding.com/sketch -->
```
petemill: Can we run all the SVGs when importing to code base through SVGO tool, or https://jakearchibald.github.io/svgomg/ UI - there are a few wasted lines here, as well as polluting ID space.

components/rewards/alert/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    return null
+  }
+
+  render () {
+    const { id, children, onClose } = this.props
+
+
+    return (
+      <StyledWrapper id={id}>
+        <StyledIcon src={this.icon} />
```
petemill: IMO we should get used to inlining icon SVG instead of treating them as images (either by using the image tag, or a background image) since this allows us to control icon styles, especially colors.

components/rewards/alert/assets/success.svg
NejcZdovc: @petemill sure I will do that. Thing is that design is not final and some icons are not final either

NejcZdovc:
As discussed even though that all components are not px perfect and there is some some stuff to do, we will merge it and do some reorganizations on the master. This way other features can start using components that were done in this PR. I will continue with improving rewards components so that they will be px perfect with the latest design (should be ready quite soon)

Per above @cezaraugusto can you please do a review

PR #43, opened 2018-07-15
More Styling

rossmoody:
Tried to be detailed in my commits. Only styling adjustments for the most part. 

Call outs worth giving more review consideration are the adjustment to the panel height and the H1 adjustments.


cezaraugusto:
please rebase

rossmoody:
In favor of #53

Issue #41, opened 2018-07-11
Create icon components


petemill:
### Goals
Icon components that share the Brave look and feel, and can be used independently or incorporated in to other Components

### Requirements
- Be able to change color of icon
- Resize icons to any size the consumer wants
- Icons are sized relative to each-other (not just alignment)
- Adds only the icons that are used to the consuming app's JS bundle

#### Not required
- Be able to change weight of icon stroke independently of icon size

### Solution option 1 (modeled on https://github.com/jacobwgillespie/styled-icons/)
- Each icon is a simple component that returns an `<svg />`.
- Each component exports shared styled-component css with common css for svg to display correctly with color / size.

#### Icon component

**./icons/arrow-right.js** _(generated from script via sketch)_
```js
const graphic = [<path d='M8.54829876 8.59872029c-.33066409-.2494677-.39648676-.71975744-.14701905-1.05042153.2494677-.33066409.71975744-.39648676 1.05042153-.14701905l5.99999996 4.52666709c.3991499.3011364.3974432.9011397-.0034134 1.2000006l-5.99999996 4.4733329c-.33207795.2475825-.80198562.1790855-1.04956815-.1529925-.24758254-.3320779-.17908548-.8019856.15299247-1.0495681l5.19828274-3.8756083-5.20169614-3.92439111z'
    />]

export const ArrowRight = styled.svg.attrs({
  children: (props) => (
    props.title != null
      ? [<title key="ArrowRight-title">{props.title}</title>, graphic]
      : [graphic]
  ),
  viewBox: '0 0 24 24',
  'aria-hidden': (props) => (props.title == null ? 'true' : undefined),
  focusable: 'false',
  role: (props) => (props.title != null ? 'img' : undefined)
})`
  width: 100%;
  height: 100%;
`
```

#### Consumer
```js
import ArrowRight from 'brave-ui/icons/arrow-right'

// Optionally give the icon a width,
// but otherwise it will take up 100% the height of the button
// and appropriate width for aspect-ratio of the graphic.
const ButtonIcon = ArrowRight.extend`
	width: 24px;
`

const Button = styled('button')`
	display: 'flex';
	flex-direction: ${p => p.iconAfterText ? 'row-reverse' : 'row'};
	font-size: 16px;
	color: 'white';
	background: 'blue';
`

export default ({ text }) => {(
	<Button>
		<ButtonIcon graphic='brave-logo' />
		{ text }
	</Button>
)}
```
[...]

PR #51, opened 2018-07-25
Adding SimpleIcon component [WIP]

ryanml:


petemill:
Hey @ryanml before we go too far down this path, did you see https://github.com/brave/brave-ui/issues/41?

I see two problems with this implementation, but please correct me if I'm wrong:

1. It doesn't avoid every single icon being bundled in to the webpack output, when an app might only need 1 icon.

2. It's passing color through as a prop. We should strive to avoid size and color specifically being passed through as props, so that they can be controlled via the css and respond to states like hover, etc. Did you see how the icons are implemented at https://github.com/brave/browser-laptop/tree/master/icons? The color can be controlled by a simple css variable.

ryanml:
Closing in favor of #54

Issue #52, opened 2018-07-26
Themeing / Global colors


petemill:
There doesn't yet seem to be anywhere to define the global color palette.

I assume this would be useful to do as a default theme, so that multiple themes can be added or overidden by a consuming app.

This should be doable without having to pass round a `theme={{ color: ... }}` prop everywhere, via a ThemeProvider as described in the first example at: https://www.styled-components.com/docs/advanced#theming

```js
// Define our button, but with the use of props.theme this time
const Button = styled.button`
  font-size: 1em;
  margin: 1em;
  padding: 0.25em 1em;
  border-radius: 3px;

  /* Color the border and text with theme.main */
  color: ${props => props.theme.main};
  border: 2px solid ${props => props.theme.main};
`;

// We're passing a default theme for Buttons that aren't wrapped in the ThemeProvider
Button.defaultProps = {
  theme: {
    main: 'palevioletred'
  }
}

// Define what props.theme will look like
const theme = {
  main: 'mediumseagreen'
};

render(
  <div>
    <Button>Normal</Button>

    <ThemeProvider theme={theme}>
      <Button>Themed</Button>
    </ThemeProvider>
  </div>
);
```

It should be ok to require that consumers of brave-ui always use a ThemeProvider and that brave-ui provides the default provider.

petemill:
After a discussion with designers, here's the preferred approach to color variables / themeing:
- Define a color palette: specific color choices that are on brand, and complimentary and we want to try to stick to., separately to the purpose of when to use the colors.
- Then define our default theme colors: which use colors from the palette where appropriate.
- We could have multiple themes - BraveLight, BraveDark, BATLight, Publishers, etc… (hopefully not all those, but this is an option for variation instead of completely different component styles).
- Palette colors can be used directly, but most of the Components will use theme colors, e.g. see below

So Components (e.g `Button`) style (e.g. `TextColor`) would use Theme (e.g. `BraveLight`) colors for specific items (e.g. `TextOnBrandColor`) which may use palette colors (e.g. `White300`).

This allows:
- A purposeful and consistent UX
- Ease of maintenance and change
- Ease of introducing theme variations (I’m pretty sure @bradrichter is keen on a dark mode for browser content areas)

Here’s an example of how this could be defined in the code, not sure how similar we are so far to this:

```
Palette
——
Red1
Red2
Red3
Purple1
Purple2
Purple3
Gray1 
Gray2
Gray3
White1
White2
White3

Theme - Brave
———
// general
[...]

PR #56, opened 2018-07-27
Make a single unified Button component (again)

petemill:
- Simplifies CSS.
- Uses <button> element for accessibility.
- Uses theme colors
- Introduces prop naming:
  - Level (primary, secondary, tertiary)
  - Type (default, accent, warn, subtle)
  - Brand (brave, rewards) _for accent type atm_
  - Size (small, medium / default, large, call-to-action)
  - disabled: bool
- Icon.position is now `'before' | 'after'` (was `'left' | 'right')

View at: https://brave-ui-opumllnkuc.now.sh


petemill:
@NejcZdovc ah so that's why - cool solution. Still, I think we can avoid the additional DOM element and transparency compositing / calculation since the latest thoughts on colors are to define the color values explicitly, especially where we do not actually want transparency - detail of that is on #52

petemill:
I've updated this PR to replace all the previous versions of Primary / Secondary / Cta Button. Ready for review after #84 (Theme).

I had to re-generate some test snapshots due to the button css changing

src/components/buttonsIndicators/button/index.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+export type Level = 'primary' | 'secondary' | 'tertiary'
+export type Type = 'default' | 'accent' | 'warn' | 'subtle'
+export type Brand = 'brave' | 'rewards'
+export type Size = 'call-to-action' | 'large' | 'medium' | 'small'
+
+export default class ThemedButton extends React.PureComponent<Props, {}> {
+  static displayName = 'Button'
+
+  static defaultProps = {
```
NejcZdovc: where is this used?

PR #63, opened 2018-08-07
Adds missing exports

NejcZdovc:
This was missed in #58.


NejcZdovc:
Merging so that new version can be published.

PR #64, opened 2018-08-07
Syncs brave-ui and brave-core

NejcZdovc:
This PR splits big settings page in stories folder into a smaller components and creates the same structure as we have it in brave-core


stories/features/rewards/settings/settingsPage.tsx
```diff
@@ -0,0 +1,59 @@
...
+            <Grant/>
+            <PageWallet/>
+          </Column>
+        </Grid>
+      </SettingsPage>
+    )
+  }
+}
+
+export default Settings
```
ryanml: Since we are importing this as `SettingsPage` let's name the export that, wdyt?

NejcZdovc: I agree, but sadly we can't because we already have component  named like this

`import { MainToggle, SettingsPage } from '../../../../src/features/rewards'`

NejcZdovc: maybe we can rename this file to settings.tsx?

ryanml: Sgtm

NejcZdovc: @ryanml done

PR #67, opened 2018-08-09
add card component

cezaraugusto:
Added card component previously called panel. there's a component in rewards with the same name so I had to adapt. I like it tho.

Ideally, the rewards components should re-use it so one-change-changes-it-all but left them as-is for now. Welcome page was updated to make use of it.

### DIFF
--- a/src/components/layout/card/style.ts
+++ b/src/components/layout/card/style.ts
@@ -1,0 +1,23 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License. v. 2.0. If a copy of the MPL was not distributed with this file.
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import styled from 'styled-components'
+import { CardProps } from './index'
+import { setTheme } from '../../../helpers'
+
+const StyledCard = styled.div`
+  max-width: ${(p: CardProps) => setTheme(p.theme, 'maxWidth')};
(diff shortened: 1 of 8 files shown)


NejcZdovc:
I agree with this. I think that we should in most cases only pass pre-defined options in the component and then component can generate style based on that. I think theming should be more about predefined options then on custom css properties. For padding and merging we can always provide wrapper in parent component that is using that component.

I tried to reduce this to minimum in rewards, but still I am passing too much px props in components. I will refactor this in near future. One good example of my abuse is Token. I am passing css for color and font size for 3 different elements. I will group this into 3 types now that I know all variations and will accept only one prop type which would accept strings small | medium | large.

src/components/layout/card/style.ts
```diff
@@ -0,0 +1,22 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License. v. 2.0. If a copy of the MPL was not distributed with this file.
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import styled from 'styled-components'
+import { CardProps } from './index'
+
+const StyledCard = styled<CardProps, 'div'>('div')`
```
NejcZdovc: nit: can we use `export const StyledCard`

NejcZdovc: because if we add one more div to it, then default will not work anymore

cezaraugusto: sure

PR #69, opened 2018-08-10
add input component

cezaraugusto:


src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,24 @@
...
+import styled from 'styled-components'
+import { Props } from './index'
+import { setTheme } from '../../../helpers'
+
+export const StyledWrapper = styled.div`
+  max-width: ${(p: Props) => setTheme(p.theme, 'maxWidth') || '254px'};
+  width: 100%;
+` as any
+
+export const StyledInput = styled.input`
```
NejcZdovc: Input should be synced with design language

src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,24 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License. v. 2.0. If a copy of the MPL was not distributed with this file.
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import styled from 'styled-components'
+import { Props } from './index'
+import { setTheme } from '../../../helpers'
+
+export const StyledWrapper = styled.div`
+  max-width: ${(p: Props) => setTheme(p.theme, 'maxWidth') || '254px'};
```
petemill: Does it need a wrapper? This Input component is mostly all 👍 from me since it's not reading css values as inline props, except here. The component can be 100% width, and if the caller wants to restrict its width, it can use its own wrapper element with a fixed width, or inside a flex / grid / whatever layout.

Or, if we want a consistent width for most form components / Input components, then have the '254px' controlled by a slightly more semantic prop, such as `fullWidth: bool`. Then it either gets the easy default, or can be put in a more flex layout by setting that prop to `true`.

NejcZdovc: I would suggest that we create something like FormWrapper, which should be used in the combination with form controls. This way we can have different wrapper types (full width, limited, etc) and controlled in one place for all controls (input, select, checkbox, radio, etc).

Something like https://github.com/brave/brave-ui/blob/62a127bd0f7bea47d547a01717009b3ecf601bf3/src/components/formControls/textarea/index.tsx#L30. But if we would use it we would need to improve ControlWrapper as you suggested, so that we are again not passing px in but type

cezaraugusto: this component is mostly a shallow copy of the current TextArea component with small variations.

we need consensus in this part I'm planning something here. Meanwhile, I'll update this per design spec so we can discuss what/what not a component should be. I'll ping you both

NejcZdovc: Yeah I think that we should pull out `ControlWrapper` from inputs/textarea and developer should wrap it in parent component.

src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,34 @@
...
+export const StyledInput = styled.input`
+  min-height: auto;
+  box-sizing: border-box;
+  width: 100%;
+  max-width: 100%;
+  font-size: 14px;
+  font-family: "Poppins", sans-serif;
+  border: 1px solid #DFDFE8;
+  border-radius: 3px;
+  padding: 12px 10px;
```
NejcZdovc: Height of the input is 2px to big, should be 40px. It's 42px currently.

src/components/formControls/input/index.tsx
```diff
@@ -0,0 +1,30 @@
...
+export default class Input extends React.PureComponent<Props, {}> {
+  render () {
+    const { id, onChange, defaultValue, disabled, placeholder } = this.props
+
+    return (
+      <StyledInput
+        id={id}
+        onChange={onChange}
+        disabled={disabled}
+        defaultValue={defaultValue}
```
NejcZdovc: this needs to be value and not default value. If it's a default value, parent component can't change it later

src/components/formControls/input/index.tsx
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react'
+import { StyledInput } from './style'
+
+export interface Props {
+  id?: string
+  defaultValue?: string
```
NejcZdovc: please change this to value

cezaraugusto:
feedback addressed now we have icons

src/components/formControls/input/style.ts
```diff
@@ -3,32 +3,50 @@
...
-  color: #686978;
-  outline: unset;
+  padding: 8px 10px;
+  opacity: ${(p: Props) => p.disabled ? '0.3' : '1'};
 
-  &:focus {
+  &:focus-within {
     border-color: #A1A8F2;
   }
+` as any
```
NejcZdovc: Can you please remove any?

More info here https://github.com/brave/brave-ui/wiki/Code-style-guide#styled-component-definition

src/components/formControls/input/index.tsx
```diff
@@ -3,28 +3,48 @@
...
 import * as React from 'react'
-import { StyledInput } from './style'
+import { InputComponent, StyledIcon, StyledInput } from './style'
 
 export interface Props {
-  id?: string
-  defaultValue?: string
+  testId?: string
+  icon?: React.ReactNode
+  type?: 'text' | 'email' | 'search' | 'password'
```
NejcZdovc: I think that we need to add text as default.

More info here https://github.com/brave/brave-ui/wiki/Code-style-guide#default-props

src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,52 @@
...
+  max-width: 100;
+  border: 1px solid #DFDFE8;
+  border-radius: 3px;
+  padding: 8px 10px;
+  opacity: ${(p: InputProps) => p.disabled ? '0.3' : '1'};
+
+  &:focus-within {
+    border-color: #A1A8F2;
+  }
+` as any
```
NejcZdovc: Can you please remove any

src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,52 @@
...
+  color: #686978;
+  outline: unset;
+
+  &::placeholder {
+    color: #B8B9C4;
+    font-weight: normal;
+    text-align: left;
+    letter-spacing: 0;
+  }
+` as any
```
NejcZdovc: Can you please remove any

cezaraugusto:
this is what I have from abstract

screen shot 2018-08-24 at 11 01 03 am

NejcZdovc:
@cezaraugusto yes I have the same in abstract. Problem is that you are doing disabled with opacity instead of doing it with colors that are defined in sketch. Because we have grey background and you use opacity input looks different

src/components/formControls/input/style.ts
```diff
@@ -0,0 +1,52 @@
...
+export const InputComponent = styled<InputProps, 'div'>('div')`
+  background-color: #fff;
+  min-height: auto;
+  box-sizing: border-box;
+  width: 100%;
+  max-width: 100;
+  border: 1px solid #DFDFE8;
+  border-radius: 3px;
+  padding: 8px 10px;
+  opacity: ${(p: InputProps) => p.disabled ? '0.3' : '1'};
```
NejcZdovc: type is not needed here, it can just be `${p =>...}`

PR #70, opened 2018-08-10
Sync page and modals concept

cezaraugusto:
demo available at https://brave-ui-lqiwxfpbin.now.sh


stories/features/sync/disabledContent.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+// Modals
+import NewToSyncModal from './modals/newToSync'
+import ExistingSyncCodeModal from './modals/existingSyncCode'
+
+// Utils
+import locale from './page/fakeLocale'
+
+interface SyncDisabledContentState {
+  newToSyncModal: boolean
```
petemill: nit: using the word 'modal' in this prop ties the implementation to the caller, who does not need to know that a modal will be used. Presumably the caller can just say that this is either 'newToSync' or 'existintSyncCode'? Further, could it just be one boolean - is it possible for both to be true or both to be false?

stories/features/sync/enabledContent.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+// Modals
+import SyncANewDeviceModal from './modals/syncNewDevice'
+import ResetSyncModal from './modals/resetSync'
+
+// Utils
+import locale from './page/fakeLocale'
+
+interface SyncEnabledContentState {
+  syncANewDeviceModal: boolean
```
petemill: Same nit re use of the word 'modal' here - stateful props passed from caller could be a little more semantic

stories/features/sync/disabledContent.tsx
cezaraugusto: it is possible for both to be `false` only, which is the initial state. Removed the "modal" word

src/features/sync/text/index.ts
```diff
@@ -0,0 +1,77 @@
...
+  margin: 20px 0;
+  padding: 0 0 0 20px;
+`
+
+export const ListOrdered = List.withComponent('ol')
+
+export const ListBullet = styled(Paragraph.withComponent('li'))`
+  margin: 10px 0;
+
+  > button {
```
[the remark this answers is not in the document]
cezaraugusto: referring to other components seems good but for our `Button` it seems to be harder than it should as TS complains about weird props such as `concat` and didn't seem worth the effort to investigate now given the timeline for this, so I ended up setting the `className`. I'll create an issue for buttons

src/components/formControls/textarea/index.tsx
```diff
@@ -8,17 +8,25 @@ import { StyledWrapper, StyledArea } from './style'
...
 
     return (
       <StyledWrapper id={id}>
-        <StyledArea onChange={onChange} disabled={disabled} value={value} />
+        <StyledArea
+          onChange={onChange}
+          disabled={disabled}
+          value={value}
+          placeholder={placeholder}
+          defaultValue={defaultValue}
```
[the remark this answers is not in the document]
cezaraugusto: ya React has it, more in https://reactjs.org/docs/uncontrolled-components.html#default-values. I ended up not using it here but seems good to have since it is valid per React code.

PR #78, opened 2018-08-16
Get a Storybook for each PR via `now` deployments

petemill:
**Please do not merge until security review passes**

This provides config so that developers can simply run the `now` command inside the repo at any time and get a permanent public Url for that version that can be shared with other team members.

The main aim though is to have `now` create a public Url for each PR so that we can easily see the Storybook for the code in that PR. I've submitted a security review for that.

I decided to use docker so that we're serving a static version of storybook, instead of the hot-reload dev server. Save the trees.

Here's an example of a deployment: https://brave-ui-roywtrxoei.now.sh

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,8 @@ node_modules
 
 # storybook
 .storybook/temp
+.storybook-out
+.out
 
 # package
 package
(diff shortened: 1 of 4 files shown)


NejcZdovc:
@petemill can you please add to wiki how to do this process, so that new devs will know how to do it

PR #81, opened 2018-08-17
Rewards5

NejcZdovc:
- removes first captcha step
- adds max-width to site banner
- add current amount prop to donate component
- add option to close site banner with esc key
- improves accessibility for all click events


src/features/rewards/grantClaim/index.tsx
```diff
@@ -10,18 +10,18 @@ const gift = require('./assets/gift')
 
 export interface Props {
   id?: string
-  onClick: () => void
+  onClaim: () => void
```
ryanml: Should this be `onClaim?:` ?

NejcZdovc: can you elaborate more why this action should be optional

ryanml: my mistake, i was thinking actions would be kept optional for now but that doesn't make sense with this one

NejcZdovc: actions are only optional when this action can't be preformed because of reason X. For example we have optional action for opening view monthly report, because we are not using this action in one other scenario

PR #84, opened 2018-08-21
Add Theme interface and first implementation for Brave-Default (light theme)

petemill:
Also adds a palette file which all (most) theme color values should come from, and any non-theme direct color values (which shouldn't happen often!)
Re-exports styled-components adding on typescript-compatible theme types at /src/theme
All components now need to be inside a <ThemeProvider> element somewhere in its ancestry. Configures Storybook to wrap all stories in it with default theme. If we have multiple themes, we could provide a switch in Storybook that changes this central location.
Created a <TestThemeProvider> with a default theme pre-configured, for use in tests.

Converts all previous non-theme `theme` props to `customStyle` (as a separate commit to aid review).

View at: https://brave-ui-wvujxeyncg.now.sh


petemill:
You might want to review the commits separately since I had to do a lot of find and replace to refactor the theme={} as well as wrap the tests

stories/components/theme.tsx
```diff
@@ -0,0 +1,114 @@
...
+// Components
+import styled from '../../src/theme'
+import paletteColors from '../../src/theme/palette'
+import themeBraveDefault from '../../src/theme/brave-default'
+import '../assets/fonts/muli.css'
+
+// wrappers
+const PaletteItemContainer = styled.div`
+  width: ${(p: any) => p.individual ? '25%' : '10%'};
+  /* min-width: 150px; */
```
NejcZdovc: can we remove this one?

src/theme/palette.ts
```diff
@@ -0,0 +1,64 @@
+// Define colors here by color name and shade, not by purpose
+const scaleGrey = {
+  grey000: '#F9F9FD',
+  grey100: '#F1F1F9',
+  grey200: '#E9E9F4',
+  grey300: '#C5C5D3',
+  grey350: '#B8B9C4',
```
NejcZdovc: didn't we say that we will remove this one

src/theme/palette.ts
```diff
@@ -0,0 +1,64 @@
...
+  grey400: '#A0A1B2',
+  grey500: '#7C7D8C',
+  grey600: '#585968',
+  grey700: '#343546',
+  grey800: '#242536',
+  grey900: '#1B1D2F'
```
NejcZdovc: and this one as well

stories/components/theme.tsx
```diff
@@ -0,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react'
+// Storybook requires
+import { storiesOf, setAddon } from '@storybook/react'
+// import { withKnobs, boolean, select } from '@storybook/addon-knobs'
```
cezaraugusto: nit: this can be removed

PR #91, opened 2018-08-22
Icon Component Generator tool and documentation

petemill:
Separate package.json to keep dependencies separate, but did not use a separate repo for now since the component template is tied to this repo.


NejcZdovc:
@petemill can we remove yarn lock as we use npm in this repo

tools/icon-component-generator/util/fixLint.js
```diff
@@ -0,0 +1,17 @@
+// const standard = require('standard')
+
+module.exports = function fixLint (data) {
+  // TODO: support ts standard as straight js standard errors
+  return Promise.resolve(data)
+  // return new Promise((resolve, reject) => standard.lintText(data, { fix: true }, (err, standardData) => {
```
cezaraugusto: do we want to keep all this commented out?

petemill: yeah I think it's ok for now this is how it was done for browser-laptop which used standard, and I don't want to lose the flow since it's not pushed in any other repo.

For now we don't need to `lint --fix` since I made the template pass this project's linter.

cezaraugusto: ok sounds good

petemill:
Removed yarn.lock and removed standard dependency

PR #94, opened 2018-08-23
Added test ids for toggle and rewards maintoggle

jasonrsadler:
Adding test-ids for rewards testing.

### DIFF
--- a/src/components/formControls/toggle/index.tsx
+++ b/src/components/formControls/toggle/index.tsx
@@ -13,6 +13,7 @@ export interface Props {
   size?: 'large' | 'small'
   type?: 'dark' | 'light'
   onToggle?: (event: React.MouseEvent<HTMLDivElement>) => void
+  testId?: string
 }
 
 export default class Toggle extends React.PureComponent<Props, {}> {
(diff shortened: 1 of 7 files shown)


jasonrsadler:
Done. Thanks.

ryanml:
@jasonrsadler can you rebase with master? (this should fix Travis) then we should be good to go

jasonrsadler:
This is still in progress.

ryanml:
@jasonrsadler why did I not see the label in front of me :D thanks

jasonrsadler:
one is a prop and one is the rendered attribute

petemill:
Should we just allow all data-* props to pass through from component to the
first rendered element of the component?

jasonrsadler:
In some cases, for my tests on `brave-core`, I need to grab an id for a toggle, which isn't available at the component from within core. 

(or maybe I didn't understand the question)

ryanml:
@jasonrsadler I mean that in `src/components/formControls/toggle/index.tsx` it is `data-test-id` as a rendered att, and in `src/features/rewards/box/index.tsx` it is `datatestid` as a rendered att.

jasonrsadler:
in `src/features/rewards/box/index.tsx` it is datatestid because it is passed as a prop down to `toggle`

jasonrsadler:
Agree. Change prop name to `testId`

PR #97, opened 2018-08-24
Adds styled components names into class names

NejcZdovc:
Before

image

After
image

You need to use this syntax to get correct naming https://github.com/brave/brave-ui/wiki/Code-style-guide#styled-component-definition


petemill:
@NejcZdovc This looks useful, but I don't understand why it's needed to get the typescript props working:
> You need to use this syntax to get correct naming brave/brave-ui/wiki/Code-style-guide#styled-component-definition

PR #98, opened 2018-08-24
Rewards 7

NejcZdovc:
- [x] use new buttons
- [x] remove any for style components 
- [x] default props
- [x] remove tooltip
- [x] add icon component
- [x] add funds
- [x] add card in box component

Missing icons
- send (Donate)
- present (GrantClaim)
- social (SiteBanner)


src/features/rewards/box/index.tsx
```diff
@@ -74,7 +73,7 @@ export default class Box extends React.PureComponent<Props, State> {
     } = this.props
 
     return (
-      <StyledWrapper id={id}>
+      <StyledCard testId={id}>
```
ryanml: Should this one have an `id` as well?

NejcZdovc: `testId` is required by `Card` component

PR #99, opened 2018-08-24
Add a span around the numeric value for Donation amount. This allows for

yachtcaptain23:
contenteditable to be used by Publishers app.


NejcZdovc:
@yachtcaptain23 can you please resolve this conflict and rebase to the latest master. Otherwise looks good to me

PR #100, opened 2018-08-26
Adding NPM script for creating a new component

ryanml:
This should hopefully make kicking off a new component a bit easier.

Baseline usage is:
`npm run create NewComponent` where `NewComponent` is whatever you'd like the React Component to be named. (Note: the default folder is `src/components`, the folder that will be made here is camel cased like `newComponent`

If you want to create the component inside features, you may run:
`npm run create NewComponent rewards` or `npm run create NewComponent newFeature`

This will create a component with a basic style, index, and spec file, as well as an assets and snapshot folder. `npm run lint` and `npm run test-unit` should pass and create snapshot respectively after running this script.

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,8 @@
     "test-unit": "jest -t",
     "test-unit-all": "jest --verbose",
     "test-security": "npm audit",
-    "coverage": "jest --coverage --verbose"
+    "coverage": "jest --coverage --verbose",
+    "create": "node ./tools/createComponent.js"
   },
   "dependencies": {
     "emptykit.css": "^1.0.1"
@@ -59,6 +60,7 @@
     "enzyme": "^3.3.0",
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-to-json": "^3.3.4",
+    "fs-extra": "^2.1.2",
     "identity-obj-proxy": "^3.0.0",
     "jest": "^23.5.0",
     "jest-styled-components": "^5.0.1",
--- a/tools/createComponent.js
+++ b/tools/createComponent.js
@@ -1,0 +1,48 @@
+const fs = require('fs-extra')
+const path = require('path')
+
+/**
+ * Usage:
+ *
+ * npm run create {ComponentName} {FeaturePath}
+ *
+ * @param {string} ComponentName - The name of the React component class to create
+ * @param {string} FeaturePath (Optional) - Defaults to src/components if not provided.
+ * This should be the name of the folder to place the component in inside src/features.
+ */
+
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -31,10 +31,12 @@
...
     "test-unit-all": "jest --verbose",
     "test-security": "npm audit",
-    "coverage": "jest --coverage --verbose"
+    "coverage": "jest --coverage --verbose",
+    "create": "node ./tools/createComponent.js"
   },
   "dependencies": {
-    "emptykit.css": "^1.0.1"
+    "emptykit.css": "^1.0.1",
+    "fs-extra": "^2.1.2"
```
NejcZdovc: Can we move this to devDependency

tools/createComponent.js
```diff
@@ -0,0 +1,38 @@
+const fs = require('fs-extra')
+const path = require('path')
+
```
NejcZdovc: Can you please add quick descriptions about arguments

tools/createComponent.js
```diff
@@ -0,0 +1,38 @@
...
+}
+
+const defaultDir = path.join(__dirname, '../src')
+const parentDir = process.argv.length === 4
+  ? path.join(defaultDir, 'features', process.argv[3])
+  : path.join(defaultDir, 'components')
+const componentDir = path.join(parentDir, getFolderName(componentName))
+
+const writeItems = [
+  {type: 'dir', path: 'assets'},
```
NejcZdovc: This will be more exception then a rule, because most of the assets will come from Icon component

tools/createComponent.js
```diff
@@ -0,0 +1,38 @@
...
+  {type: 'dir', path: '__snapshots__'},
```
NejcZdovc: I think this can be removed, because it's automatically generated

ryanml:
@NejcZdovc comments addressed

tools/lib/templates.js
```diff
@@ -0,0 +1,79 @@
...
+module.exports.style = () => `\
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this file,
+* You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+import styled from 'styled-components'
+
+export const StyledWrapper = styled<{}, 'div'>('div')\`
+  width: flex;
```
NejcZdovc: should be `display: flex`

ryanml: damn you autocomplete... will fix :)

PR #111, opened 2018-08-28
Removes unused fonts

NejcZdovc:


NejcZdovc:
closing as wontfix for now. We will keep all of them

PR #114, opened 2018-08-29
Using heading component for rewards

### LINKED ISSUE #105: Use heading component in rewards

ryanml:
Fixes #105

### DIFF
--- a/src/features/rewards/welcomePage/index.tsx
+++ b/src/features/rewards/welcomePage/index.tsx
@@ -25,8 +25,8 @@ import {
   StyledTakeActionContent,
   StyledBackground,
   StyledBatLogo,
-  StyledH1,
-  StyledH2,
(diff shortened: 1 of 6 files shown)


src/features/rewards/welcomePage/index.tsx
```diff
@@ -83,11 +83,11 @@ class WelcomePage extends React.PureComponent<Props, {}> {
           <StyledBatLogo>
             <BatColorIcon />
           </StyledBatLogo>
-          <StyledH1>
+          <StyledRewardsTitle level={1}>
```
NejcZdovc: Should be level 2. Because with level 1 we change it from 38px to 48px. Let's use level 2 which is 40px so that jump is not that huge

ryanml: Sounds good

Issue #113, opened 2018-08-28
Profile component does not quite match design spec


ryanml:

ryanml:
Closed in #120

PR #120, opened 2018-08-31
Wallet Panel Concept Improvements

ryanml:
1. Adds the gradient background shown in `WalletSummary` per abstract
2. Fixes profile per #113 

Profile Before:
screen shot 2018-09-04 at 11 48 09 am

Profile After:
screen shot 2018-09-04 at 4 08 49 pm


src/features/rewards/walletSummary/index.tsx
```diff
@@ -31,105 +32,124 @@ export interface Props {
   total: Token
   id?: string
   compact?: boolean
+  gradient?: string
```
NejcZdovc: Sorry, didn't saw it before. Can we do the same here with `compact`

PR #150, opened 2018-09-12
let onRestore prop be optional in modalContribute

### LINKED ISSUE #149: unable to run npm run `storybook-deploy`
ModalContribute prop `onRestore` is not set and makes `npm run storybook-deploy` unable to run.

If `onRestore` is optional it should be flagged this way for TypeScript

cezaraugusto:
fix #149

component interface says that `onRestore` is required (see [here](https://github.com/brave/brave-ui/blob/master/src/features/rewards/modalContribute/index.tsx#L21)) but implementation doesn't use it (see [here](https://github.com/brave/brave-ui/blob/master/stories/features/rewards/settings/contributeBox.tsx#L169)).

### DIFF
--- a/src/features/rewards/modalContribute/index.tsx
+++ b/src/features/rewards/modalContribute/index.tsx
@@ -18,7 +18,7 @@ import { getLocale } from '../../../helpers'
 export interface Props {
   rows: DetailRow[]
   onClose: () => void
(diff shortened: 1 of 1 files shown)


cezaraugusto:
#147 fixes this so I'm going to close

Issue #122, opened 2018-08-31
Use tooltip


NejcZdovc:
Add tooltip in rewards pages where we have icon, but no label that would explain this icon. You can see examples in abstract.

PR #154, opened 2018-09-18
Adds exclusion tooltip, locale fix

ryanml:
This finishes off #122 and fixes an incorrect localized string


src/features/rewards/tooltip/style.ts
```diff
@@ -6,23 +6,24 @@
 import styled from 'styled-components'
 
 interface StyleProps {
+  left?: string
```
NejcZdovc: Can we instead of adding left prop, add `display: inline-block` to `StyledWrapper`. This way wrapper will take width of the child element and it will be always 50%?

ryanml: if this prop is not present it will default to `inline-block` at `50%` left. in the case of the site table, the tooltip with display as `inline-block` and at `50%` left is off, as it's pointing to the middle of the contribute table cell, and not the `x` button.

If not left, I think we should have some way to pass in a custom adjustment.

src/features/rewards/tableContribute/style.ts
```diff
@@ -13,7 +13,6 @@ export const StyledText = styled<{}, 'div'>('div')`
 `
 
 export const StyledRemove = styled<{}, 'button'>('button')`
-  margin: 0 5px 0 0;
```
ryanml: This margin was superfluous and causing issues with the tooltip alignment

Issue #126, opened 2018-09-05
Select overflow


NejcZdovc:
Select is currently not overflowing all elements.

![image](https://user-images.githubusercontent.com/9574457/45073349-3fd63a80-b0df-11e8-986d-8ce2446c1d7c.png)

PR #160, opened 2018-09-24
Adds single panel option to the slider

NejcZdovc:


jasonrsadler:
#126 will apply here as well.

Issue #174, opened 2018-10-02
Add Radio Button to Shared Components


ryanml:
Add a radio button implementation under form control for shared components

For the label, it will accept either a string which will then be styled, or react node children which will be left to their own styling devices.

PR #177, opened 2018-10-03
Adding Radio button component

ryanml:
Fixes: #174 

Concept shown here:

screen shot 2018-10-03 at 10 22 47 am

**Note: As Radio buttons are stilled marked WIP in the design spec, a pixel perfect match has not been done with this concept. It's pretty close but this will unblock us on adding the mobile selection option for android UI.**

### DIFF
--- a/src/components/formControls/radio/style.ts
+++ b/src/components/formControls/radio/style.ts
@@ -1,0 +1,94 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License. v. 2.0. If a copy of the MPL was not distributed with this file.
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import styled, { css, ThemedStyledProps } from '../../../theme'
+import { Size } from './index'
+
+interface StyleProps {
+  size?: Size
+  disabled?: boolean
+  selected?: boolean
+}
+
+const getThemeSizes = (p: ThemedStyledProps<StyleProps>) => {
+  let fillSize = '8'
+  let circleSize = '18'
+  let fontSize = '14'
+  let rightMargin = '12'
+
+  if (p.size === 'big') {
+    fillSize = '10'
+    circleSize = '24'
+    fontSize = '16'
+    rightMargin = '17'
+  }
+
+  return css`
+    --radio-fill-size: ${fillSize}px;
+    --radio-circle-size: ${circleSize}px;
+    --radio-font-size: ${fontSize}px;
+    --radio-right-margin: ${rightMargin}px;
+  `
+}
+
+const getThemeColors = (p: ThemedStyledProps<StyleProps>, selected: boolean | undefined) => {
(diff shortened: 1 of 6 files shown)


src/components/formControls/radio/style.ts
```diff
@@ -0,0 +1,70 @@
...
+  line-height: 1.3;
+  display: flex;
+  margin-bottom: 20px;
+  color: #686978;
+  font-size: ${p => p.size === 'big' ? '16' : '14'}px;
+  cursor: pointer;
+`
+
+export const StyledInput = styled<{}, 'input'>('input')`
+  display: none;
```
petemill: It may be better to hide this via `opacity: 0; position: absolute;` so that it's still present, for accessibility

src/components/formControls/radio/style.ts
```diff
@@ -0,0 +1,70 @@
...
+  border: 1px solid var(--radio-border-color);
+  margin-right: ${p => p.size === 'big' ? '17' : '12'}px;
+  border-radius: 50%;
+`
+
+export const StyledFill = styled<StyleProps, 'div'>('div')`
+  width: ${p => p.size === 'big' ? '10' : '8'}px;
+  height: ${p => p.size === 'big' ? '10' : '8'}px;
+  border-radius: 50%;
+  ${p => getThemeColors(p as ThemedStyledProps<Props>, p.selected)}
```
petemill: May be cleaner to put this function call at the top of this set of rules. Also is the `as ThemeStyledProps<Props>` neccessary? `p` is `ThemedStyleProps<StyleProps>` already - can't the function use that instead?

src/components/formControls/radio/style.ts
```diff
@@ -0,0 +1,70 @@
...
+import { Props, Size } from './index'
+
+interface StyleProps {
+  size?: Size
+  disabled?: boolean
+  selected?: boolean
+}
+
+const getThemeColors = (p: ThemedStyledProps<Props>, selected: boolean | undefined) => {
+  let fillColor = '#696FDC'
```
petemill: Would be great for all colors to come from theme. Can you work with a designer to make sure the color is:
1. From the standard palette
2. Has a semantic name that's appropriate across-controls if applicable, or radio-specific if not.

This is so we can do dark mode if neccessary.

At the very least, if this color would never ever change across themes, then this can come as a variable from palette.js

src/components/formControls/radio/style.ts
ryanml: Moved to top of function and `as` typing removed (I had to move some things around to not need to do that conversion)

src/components/formControls/radio/style.ts
ryanml: Done, met with @karenkliu and got the correct fill color from the palette.

ryanml:
Sounds good, thanks @petemill 🎉

PR #181, opened 2018-10-04
Loader component

rossmoody:
In all it's glory, here is the icon component loader. A color is designated within the svg which I'm unsure is the right approach since this may need to be white sometimes on buttons but thinking maybe this could be targeted with css. Either way it's a start.


NejcZdovc:
@rossmoody can you please fix linter. Thx

rossmoody:
Think i got it @NejcZdovc

NejcZdovc:
@rossmoody looks good. Could you please squash commits into one?

PR #183, opened 2018-10-04
Hide grant button if no grants

NejcZdovc:


NejcZdovc:
sure, sounds good. nice idea. will add it tomorrow

NejcZdovc:
@ryanml done

PR #184, opened 2018-10-04
Select options use modal with radio selection on touchscreen devices

### LINKED ISSUE #175: Select Options should display in modal for touchscreen devices
On Touchscreen devices, options should display in a modal with radio button selection like so:

screen shot 2018-10-02 at 3 38 06 pm

ryanml:
Design: 

Fixes #175 

screen shot 2018-10-04 at 12 07 30 pm

This can be tested by viewing the `Select` concept in the storybook, going in to full screen mode, and using the mobile device emulator in dev tools (you will have to reload the page inside the emulator)

It should adhere to design and do the following:

1. Selecting an option in the modal should change the dropdown as expected
2. Selecting a new option will dismiss the modal
3. Clicking outside the modal will also dismiss the modal

Note that this adds an optional `title` prop to `<Select />`. This is so the field title can be passed to the options modal.


src/components/formControls/select/index.tsx
```diff
@@ -93,6 +106,72 @@ export default class Select extends React.PureComponent<Props, State> {
...
+  onRadioOptionClick = (key: string, selected: boolean, child: React.ReactNode, all: {[key: string]: boolean}) => {
+    this.setState({
+      value: key,
+      selected: child,
+      radioShown: false
+    })
+  }
+
+  onOverlayClick = (event: any) => {
+    if (event.target.className.includes('StyledOptionsOverlay')) {
```
petemill: @ryanml Is this a styled-components class name? Will this string still be present in production build mode? Is there a better way to ensure this click handler is only fired by that target? Or perhaps add and read a data-* attribute to the element

ryanml: @petemill good catch, I'll adjust this to use a data-* attribute 👍

ryanml: @petemill done

src/components/formControls/select/style.ts
```diff
@@ -122,3 +122,45 @@ export const StyledOptionText = styled<StyleProps, 'div'>('div')`
...
+  background: rgba(12,13,33,0.85);
+  align-items: center;
+  z-index: 999;
+  justify-content: center;
+`
+
+export const StyledOptionsModal = styled<StyleProps, 'div'>('div')`
+  border-radius: 7px;
+  background: #fff;
+  height: 332px;
```
petemill: The fixed height here is a bit peculiar to me. Is this a design decision? On larger screens it seems a lot of space is wasted and the modal uneccessarily scrolls:
image

src/components/formControls/select/style.ts
```diff
@@ -122,3 +122,45 @@ export const StyledOptionText = styled<StyleProps, 'div'>('div')`
...
+  width: 295px;
+  z-index: 9999;
+  padding: 42px 55px 30px;
+`
+
+export const StyledSelectTitle = styled<StyleProps, 'span'>('span')`
+  color: #4B4C5C;
```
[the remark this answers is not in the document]
ryanml: I will double check and update if so 👍

src/components/formControls/select/style.ts
ryanml: I had the design in abstract to go off of, which just showed the modal concept on a smaller smartphone viewport, so I lifted the values from that.

We should probably do a percentage for width/height and aim for the modal to take up a certain portion of the window. wdyt? cc: @jenn-rhim

src/components/formControls/select/style.ts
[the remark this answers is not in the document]
ryanml: @petemill this now uses a color from the theme

src/components/formControls/select/style.ts
ryanml: @petemill 
The height is now a variable percentage to accommodate different viewports, iPad example:

screen shot 2018-10-09 at 9 13 09 am

PR #186, opened 2018-10-07
Add loader to welcome page

NejcZdovc:
check that when you are on welcome page and if you click on a button you get loader with creating wallet text button


NejcZdovc:
@jasonrsadler added disabled flag

PR #190, opened 2018-10-08
CSS Tweaks and Locale Text

rossmoody:
- updated text strings for buttons and paragraphs. few grammatical changes
- general aesthetic tweaks to typography and coloring
- updated 3 icons that were not rendering correctly
- updated input type to be body copy


rossmoody:
sync

PR #192, opened 2018-10-10
Adds notification support to walletWrapper

### LINKED ISSUE #191: Add notification support to Wallet Panel

ryanml:
Fixes #191 

Can be viewed in the wallet panel concept, here are screenshots of the implementation for the 3 notification types:

screen shot 2018-10-09 at 6 14 26 pm

screen shot 2018-10-09 at 6 15 18 pm

screen shot 2018-10-09 at 6 16 01 pm


src/features/rewards/walletWrapper/index.tsx
```diff
@@ -71,18 +92,21 @@ export interface Props {
...
   id?: string
-  gradientTop?: string
+  gradientTop?: string,
+  notification?: Notification
 }
 
 interface State {
-  grantDetails: boolean
+  grantDetails: boolean,
+  notificationDismissed: boolean
```
NejcZdovc: We don't need dismiss state. We should only have onNotificationDismissed callback and that will be handled by client.

ryanml: @NejcZdovc removed from state, now uses callback

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -71,7 +92,9 @@ export interface Props {
   grants?: Grant[]
   alert?: AlertWallet | null
   id?: string
-  gradientTop?: string
+  gradientTop?: string,
+  notification?: Notification,
+  onCloseNotification?: () => void
```
NejcZdovc: can be this part of `Notification` interface so that everything about notification is grouped

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -113,6 +163,62 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
     return grants && grants.length > 0
   }
 
+  hasNotifications = (notifications?: Notification[]) => {
```
NejcZdovc: I think this is not used

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -113,6 +163,62 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
...
+        icon = <MegaphoneIcon />
+        break
+      case 'contribute':
+        icon = <LoveIcon />
+        break
+      case 'token':
+        icon = <GiftIcon />
+        break
+      default:
+        icon = ''
```
NejcZdovc: in this case I think case we should return null. This way we don't have 58x58 big empty block

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -105,6 +128,33 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
     })
   }
 
+  generateNotification = (notification: Notification | undefined) => {
+    if (!notification) {
+      return null
+    }
+
+    return (
+      <>
```
NejcZdovc: this one is not needed

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -138,73 +245,79 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
         <StyledWrapper
           id={id}
           compact={compact}
+          notificationShown={notification}
         >
           <StyledHeader>
             {
-              alert && alert.node
-              ? <StyledAlertWrapper>
+              !notification
```
NejcZdovc: Can we do it like we are doing alert, so that we don't need this big if statement

src/features/rewards/walletWrapper/style.ts
```diff
@@ -2,13 +2,16 @@
...
+import { Notification, NotificationType } from './'
 import styled from 'styled-components'
 
 interface StyledProps {
   connected?: boolean,
   contentPadding?: boolean,
   compact?: boolean,
-  background?: string
+  background?: string,
+  type?: NotificationType,
```
NejcZdovc: can we have just one type for notification

```
notification?: Notification | undefined
``` 

type is already part of `Notification` object

src/features/rewards/walletWrapper/index.tsx
NejcZdovc: also this needs to change to `onCloseNotification?: (id: string) => void`, so that we know which one we are closing

src/features/rewards/walletWrapper/index.tsx
ryanml: Having the notification present isn't as simple as "Is there a notification present? If so, display element, if not, hide it" because I am using the `StyledWrapper` just the same and just changing the inner content for the notification.

PR #207, opened 2018-10-15
Fix types

noelyoo:
This PR fixes the type of Amount props in `src/features/rewards/amount/index.tsx`: 
* amount: `string` -> `number`
* converted: `string` -> `number`

### DIFF
--- a/src/features/rewards/amount/index.tsx
+++ b/src/features/rewards/amount/index.tsx
@@ -10,7 +10,7 @@ import { BatColorIcon } from '../../../components/icons'
 export interface Props {
   amount: string
   converted: string
-  onSelect: (amount: number) => void
+  onSelect: (amount: string) => void
   id?: string
   selected?: boolean
   type?: 'big' | 'small'
@@ -21,7 +21,7 @@ export default class Amount extends React.PureComponent<Props, {}> {
   static defaultProps = {
     type: 'small',
     currency: 'USD',
-    converted: 0
+    converted: '0.00'
   }
 
   render () {
--- a/src/features/rewards/amount/spec.tsx
+++ b/src/features/rewards/amount/spec.tsx
@@ -6,7 +6,7 @@ import Amount from './index'
 import { TestThemeProvider } from '../../../theme'
 
 describe('Amount tests', () => {
-  const baseComponent = (props?: object) => <TestThemeProvider><Amount id='amount' amount={1} converted={0.4} onSelect={() => {}} {...props}  /></TestThemeProvider>
+  const baseComponent = (props?: object) => <TestThemeProvider><Amount id='amount' amount={'1'} converted={'0.4'} onSelect={() => {}} {...props} /></TestThemeProvider>
 
   describe('basic tests', () => {
     it('matches the snapshot', () => {


NejcZdovc:
@noelyoo hi thank you for the PR. We convert all number to string in client that's why we have string for amounts

src/features/rewards/amount/index.tsx
```diff
@@ -1,41 +1,41 @@
...
+
+import * as React from 'react'
+import { StyledWrapper, StyledAmount, StyledNumber, StyledLogo, StyledConverted, StyledTokens } from './style'
+import { getLocale } from '../../../helpers'
+import { BatColorIcon } from '../../../components/icons'
+
+export interface Props {
+  amount: number
+  converted: number
+  onSelect: (amount: number) => void
```
noelyoo: @NejcZdovc Hi, then how about this `onSelect: (amount: number) => void`?. 
It says the type of `amount` is number. Should it be string too?

NejcZdovc: yes, we should be using string everywhere

src/features/rewards/amount/index.tsx
```diff
@@ -1,41 +1,41 @@
...
+  selected?: boolean
+  type?: 'big' | 'small'
+  currency?: string
+}
+
+export default class Amount extends React.PureComponent<Props, {}> {
+  static defaultProps = {
+    type: 'small',
+    currency: 'USD',
+    converted: 0
```
NejcZdovc: we need to set this to string `'0.00'` as well

src/features/rewards/amount/index.tsx
noelyoo: Right, then I would like to change it to string. Would it be ok?

NejcZdovc: that would be perfect, thank you

src/features/rewards/amount/index.tsx
noelyoo: Got it

src/features/rewards/amount/spec.tsx
```diff
@@ -6,7 +6,7 @@ import Amount from './index'
 import { TestThemeProvider } from '../../../theme'
 
 describe('Amount tests', () => {
-  const baseComponent = (props?: object) => <TestThemeProvider><Amount id='amount' amount={1} converted={0.4} onSelect={() => {}} {...props}  /></TestThemeProvider>
+  const baseComponent = (props?: object) => <TestThemeProvider><Amount id='amount' amount={'1'} converted={'0.4'} onSelect={() => {}} {...props} /></TestThemeProvider>
```
noelyoo: I also made changes here for the testing. Types should be clear now :)

NejcZdovc: nice find

NejcZdovc:
@noelyoo one small thing. Not sure, but why is index.tsx all highlighted and not just two lines. Did you maybe changed EOL?

noelyoo:
@NejcZdovc It was because of `^M`

I'm using `macOS 10.14 (18A391)` and not sure why those `^M`s got in the only `index.tsx`, not `spec.tsx`.

I fixed them right. Thanks.

noelyoo:
@NejcZdovc it turned out that I had `end_of_line = crlf` in my `.editorconfig` at the moment.

NejcZdovc:
@cezaraugusto this change looks good, but this will break some other code where `onSelect` expects number

noelyoo:
True, this PR needs more changes from other classes, i.e.`Donate` Class. This PR is incomplete.

noelyoo:
True, This PR is incomplete. There should be more changes in other classes, e.g. `Donate` Class.

noelyoo:
I will close this PR and reopen it when ready.

PR #219, opened 2018-10-17
Welcome page tweaks

ryanml:
This includes a view small design tweaks targeted for mobile per design review:

1. Reduces the font size of the hero header/subtitle on smaller devices
2. On touchscreen devices removes the outer settings page wrapper/desktop header
3. Reduces hero padding for smaller devices

### DIFF
--- a/src/features/rewards/hero/index.tsx
+++ b/src/features/rewards/hero/index.tsx
@@ -7,15 +7,19 @@ import { StyledHero } from './style'
 
 export interface Props {
   id?: string
+  isMobile?: boolean
   children?: React.ReactNode
 }
 
 export default class Hero extends React.PureComponent<Props, {}> {
   render () {
-    const { id, children } = this.props
+    const { id, isMobile, children } = this.props
 
     return (
-      <StyledHero id={id}>
+      <StyledHero
+        id={id}
+        isMobile={isMobile}
+      >
         {children}
       </StyledHero>
     )
--- a/src/features/rewards/welcomePage/index.tsx
+++ b/src/features/rewards/welcomePage/index.tsx
@@ -55,11 +55,13 @@ export interface Props {
 }
 
 class WelcomePage extends React.PureComponent<Props, {}> {
+  private isTouchScreen: boolean
   private centerTextSection: HTMLDivElement | null
 
   constructor (props: Props) {
     super(props)
     this.centerTextSection = null
+    this.isTouchScreen = 'ontouchstart' in document.documentElement
   }
 
   scrollToCenter = () => {
@@ -86,7 +88,10 @@ class WelcomePage extends React.PureComponent<Props, {}> {
 
   hero () {
     return (
-      <Hero id={'rewards-hero'}>
+      <Hero
+        id={'rewards-hero'}
+        isMobile={this.isTouchScreen}
+      >
         <StyledSection>
           <StyledBatLogo>
             <BatColorIcon />
@@ -212,13 +217,11 @@ class WelcomePage extends React.PureComponent<Props, {}> {
     ]
   }
 
-  render () {
-    const { id, onReTry } = this.props
-
+  get welcomePageContent () {
     return (
-      <SettingsPage id={id}>
+      <>
         {
-          onReTry
+          this.props.onReTry
           ? <StyledAlert>
             <Alert type={'error'}>
               <StyledAlertContent>
@@ -229,7 +232,7 @@ class WelcomePage extends React.PureComponent<Props, {}> {
                     level={'primary'}
                     type={'accent'}
                     text={getLocale('walletFailedButton')}
-                    onClick={onReTry}
+                    onClick={this.props.onReTry}
                   />
               </StyledAlertContent>
             </Alert>
@@ -255,6 +258,21 @@ class WelcomePage extends React.PureComponent<Props, {}> {
             </StyledTakeActionContent>
           </StyledCenterSection>
         </StyledBackground>
+      </>
+    )
+  }
+
+  render () {
+    const { id } = this.props
+
+    // We don't need the SettingsPage wrapper on touchscreen devices
+    if (this.isTouchScreen) {
+      return this.welcomePageContent
+    }
+
+    return (
+      <SettingsPage id={id}>
+        {this.welcomePageContent}
       </SettingsPage>
     )
   }
--- a/src/features/rewards/welcomePage/style.ts
+++ b/src/features/rewards/welcomePage/style.ts
@@ -10,6 +10,11 @@ const centerBackground = require('./assets/centerTextBackground.svg')
 export const StyledOptInSection = styled<{}, 'section'>('section')`
   margin: 40px auto;
   max-width: 303px;
+
+  @media (max-width: 410px) {
+    margin: 40px 20px;
+    max-width: unset;
+  }
 `
 
 export const StyledOptInSecond = styled<{}, 'section'>('section')`
@@ -71,6 +76,10 @@ export const StyledBackground = styled<{}, 'div'>('div')`
 export const StyledBatLogo = styled<{}, 'div'>('div')`
   margin: 5px auto 0;
   height: 152px;
+
+  @media (max-width: 460px) {
+    height: 100px;
+  }
 `
 
 export const StyledRewardsTitle = styled(Heading)`
@@ -78,6 +87,10 @@ export const StyledRewardsTitle = styled(Heading)`
   color: #FFF;
   display: inline-block;
   margin: 17px 0 4px;
+
+  @media (max-width: 360px) {
+    font-size: 36px;
+  }
 `
 
 export const StyledActionTitle = styled(Heading)`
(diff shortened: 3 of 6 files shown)


src/components/buttonsIndicators/button/index.tsx
```diff
@@ -21,6 +21,8 @@ export interface Props {
   onClick?: () => void
   id?: string
   disabled?: boolean
+  color?: string
+  heavy?: boolean
```
ryanml: These additions were requested by @jenn-rhim 

These are simple, non-breaking overrides that let you specify a certain color for a button regardless of its Type, and to explicitly use `font-weight: 700` for an extra bold texted button

NejcZdovc: cc @petemill because this is completely against what we defined in language and on our meetings

ryanml: I think this warrants a discussion for the next front-end component meeting:

We have this design goal for the Welcome Page for example, and this includes certain weights and color changes that violate what is in our standard button component. I am in between a rock and a hard place trying to both respect the shared components and implement the product as delivered. It will be good to be on the same page with this.

NejcZdovc: totally agree, design and development need to be in sync with this one, on the same page. that means everybody

ryanml: In the mean time I am going to revert the Button related portions of this PR, and we can review based on that. A follow up PR will be produced if necessary as the result of the above meeting.

jenn-rhim: We need to add 'white' buttons, both solid and outlined, in the library for the cases where the buttons are placed on the (dark) colored background. We already have many cases in production using this style that we can benefit.

NejcZdovc: then we need to just add it to design language and then add it to shared component as predefined type. As soon as we will add option for custom color this will be abused IMHO

petemill: Hey @jenn-rhim @ryanml - good news: we have a proposed plan for this that hopefully solves all the requirements, and probably needs re-presenting. One of the reasons we don't want to just throw around arbitrary colors (aside from maintaining consistency and change-ability), is still supporting 'themes' e.g. for dark mode. Bad news: the plan has not yet been implemented because we're yet to be able to get buy-in from everyone and adjust the process to use theme properties properly. I'm not sure I'll do the plan justice in this comment, but I'll attempt it:

Right now we have Button variations that are defined by semantic names, that result in different colors: 
image

The problem described here is that sometimes these buttons are on dark background and sometimes are on light background. Obviously the Default Button (which currently results in a Black button) won't look good on a dark background. The design requirement is that dark backgrounds would want different color Buttons for the same states. Right now, these darker backgrounds are used in specific known places. However, now / soon, the user will be able to make _some_ currently-light backgrounds in to dark backgrounds by changing theme preference.

So the plan is to have a concept of _theme containers_, that will know whether they are dark or light, some are always dark or light, and some will be changed as a result of user theme choice (dark or light). Essentially each part of a layout would be one of 3 things as defined by a designer:
- Always Light
- Always Dark
- Light / Dark depending on preference
(and of course the container itself would be able to have arbitrary and custom background color / graphic / whatever)

In these containers, the theme will use either Brave-Light values or Brave-Dark values depending on which theme the container is set to.

Therefore a 'Default' Button will be Black in a light container and White in a dark container. The Button would still be defined semanticly and wouldn't be able to be 'any' color over the variations that the team has defined:

| Colors        | Light-mode value           | Dark-mode value  |
| ------------- |:-------------:| -----:|
|    brandBraveInteracting | colors.orange500  |  ??? |
|    brandBraveActive | colors.orange200  |  ??? |
|    brandBraveLight | colors.orange000  |  ??? |
|     |
|    brandBatInteracting | colors.blurple500  |  ??? |
|    brandBatActive | colors.blurple200  |  ??? |
|     |
|    defaultControl | colors.grey800  | ??? |
|    defaultControlInteracting | colors.black  |  ??? |
|    defaultControlActive | colors.grey700  |  ??? |
|     |
|    warn | colors.red500  |  ??? |
|    warnInteracting | colors.red600  |  ??? |
|    warnActive | colors.red700  |  ??? |
|     |
|    subtle | colors.grey400  |  ??? |
|    subtleInteracting | colors.grey500  |  ??? |
|    subtleActive | colors.grey600  |  ??? |
    

This probably needs more of a discussion, and if this is an urgent requirement, we can hard-code a color for now, but it'd be great to ratify a solution similar to the one I've described here soon. Especially since we should be quite protective of the Button as it's an example of the rules we want to implement in other controls that allow us to reach the goals of the design system.

src/components/buttonsIndicators/button/style.ts
```diff
@@ -116,7 +117,7 @@ export const StyledText = styled<Props, 'div'>('div')`
   align-items: center;
   text-align: center;
   letter-spacing: 0;
-  font-weight: 500;
+  font-weight: ${p => p.heavy ? '700' : '500'};
```
petemill: @jenn-rhim @karenkliu what semantically makes a Button get heavy text? Should all Primary action buttons be heavy for example? If it's more arbitrary than that, can we get a 1-sentence rule describing the reason it's used for this 1 instance here and perhaps something semantic will come out of that?

ryanml:
@petemill @NejcZdovc @jenn-rhim there is some great discussion on here pertaining to the button issue (one that will be fully resolved in tomorrows meeting I hope), but the important thing in this PR is some of the spacing updates to the welcome page.

I'm going to revert the button changes from this PR, and then once we are all on the same page, we can do Button component updates in its own PR with own issue. For now we can just look at the other welcome page updates.

cc: @anthonypkeane @gdregalo

src/features/rewards/welcomePage/index.tsx
```diff
@@ -84,9 +84,16 @@ class WelcomePage extends React.PureComponent<Props, {}> {
     this.props.optInAction()
   }
 
+  isTouchscreen = () => {
+    return 'ontouchstart' in document.documentElement
```
petemill: this doesn't change, so can be a const at the module level, save re-evaluating (admittedly very very quick re-evaluation!)

src/features/rewards/welcomePage/style.ts
```diff
@@ -71,13 +76,21 @@ export const StyledBackground = styled<{}, 'div'>('div')`
...
 `
 
 export const StyledRewardsTitle = styled(Heading)`
   font-weight: 500;
   color: #FFF;
   display: inline-block;
   margin: 17px 0 4px;
+
+  @media (max-width: 360px) {
+    font-size: 36px;
```
petemill: Should this responsive behavior be built in to Heading?

src/features/rewards/welcomePage/style.ts
```diff
@@ -102,6 +115,10 @@ export const StyledSubTitle = styled(Heading)`
   text-align: center;
   line-height: 28px;
   margin: 18px 0 7px;
+
+  @media (max-width: 360px) {
+    font-size: 22px;
```
petemill: Same re: Heading. Otherwise is there much point in using the Heading component if we're setting the font-size here?

src/features/rewards/welcomePage/index.tsx
ryanml: Done

ryanml:
@petemill now using custom StyledComponents for the headers in question, as adding media queries at the Heading module level may have unintended results elsewhere

src/features/rewards/hero/index.tsx
```diff
@@ -7,15 +7,19 @@ import { StyledHero } from './style'
 
 export interface Props {
   id?: string
+  roundBorder?: boolean
```
NejcZdovc: I think that we should be using isMobile or something like this flag, so that we can use it for other changes as well if needed

src/features/rewards/welcomePage/style.ts
```diff
@@ -71,13 +77,24 @@ export const StyledBackground = styled<{}, 'div'>('div')`
...
   margin: 5px auto 0;
   height: 152px;
+
+  @media (max-width: 460px) {
+    height: 100px;
+  }
 `
 
-export const StyledRewardsTitle = styled(Heading)`
+export const StyledRewardsTitle = styled<{}, 'span'>('span')`
```
NejcZdovc: Why was this changed from Heading?

src/features/rewards/welcomePage/style.ts
```diff
@@ -96,12 +113,19 @@ export const StyledCenterTitle = styled(Heading)`
   margin: 0;
 `
 
-export const StyledSubTitle = styled(Heading)`
+export const StyledSubTitle = styled<{}, 'span'>('span')`
```
NejcZdovc: Same as above

src/features/rewards/welcomePage/style.ts
ryanml: I had discussed this with @petemill 

The font size for a couple of the items in the welcome header needs to change based on screen size. Adding this media query override kind of defeats the purpose of using the `Heading` component (which shouldn't be changed). Adding this change to the `Heading` component style could have potentially unrelated consequences, so this seemed the easiest thing to do.

src/features/rewards/hero/index.tsx
ryanml: Sure, can change

src/features/rewards/welcomePage/style.ts
NejcZdovc: you can still use header component like was before and just add media query. This way you use standard component and have custom mobile

PR #222, opened 2018-10-18
Adds wallet failed flow

NejcZdovc:


NejcZdovc:
merging, because we need this in 0.55. If you find any problems with it please comment on this PR and I will do follow up

Issue #224, opened 2018-10-19
Long publisher titles overflow in wallet panel


ryanml:
Problem:

<img width="480" alt="screen shot 2018-10-18 at 8 06 32 pm" src="https://user-images.githubusercontent.com/8732757/47195613-7bd10200-d311-11e8-9766-bc26bfb51d87.png">

Should behave like:

<img width="433" alt="screen shot 2018-10-18 at 8 08 13 pm" src="https://user-images.githubusercontent.com/8732757/47195635-8f7c6880-d311-11e8-9383-5cf87c98d0ff.png">

PR #225, opened 2018-10-19
Fixes long publisher titles breaking out of wallet panel

ryanml:
Fixes: #224 

Fix pictured:
screen shot 2018-10-18 at 8 44 15 pm

screen shot 2018-10-18 at 8 37 11 pm

screen shot 2018-10-18 at 8 36 46 pm

### DIFF
--- a/src/features/rewards/profile/style.ts
+++ b/src/features/rewards/profile/style.ts
@@ -57,13 +57,12 @@ export const StyledContent = styled<Partial<Props>, 'div'>('div')`
 `
 
 export const StyledTitleWrap = styled<Partial<Props>, 'div'>('div')`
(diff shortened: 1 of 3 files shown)


src/features/rewards/profile/style.ts
```diff
@@ -57,13 +57,12 @@ export const StyledContent = styled<Partial<Props>, 'div'>('div')`
 `
 
 export const StyledTitleWrap = styled<Partial<Props>, 'div'>('div')`
+  width: 280px;
```
NejcZdovc: I always prefer to use max-width then width. This allows responsive design in case we want to resize panel

PR #231, opened 2018-10-23
Enforces 1024px min-width for settingsPage

### LINKED ISSUE #230: Desktop settings page should have defined min width
Min width should be set to 1024px, so content does not improperly break down on window resize

See: https://github.com/brave/brave-browser/issues/1556

ryanml:
Fixes #230 

Related: https://github.com/brave/brave-browser/issues/1556

Before:
screen shot 2018-10-22 at 6 15 35 pm

After: 
screen shot 2018-10-22 at 6 15 57 pm

### DIFF
--- a/src/features/rewards/welcomePage/__snapshots__/spec.tsx.snap
+++ b/src/features/rewards/welcomePage/__snapshots__/spec.tsx.snap
@@ -15,7 +15,7 @@ exports[`WelcomePage tests basic tests matches the snapshot 1`] = `
 .c0 {
   background: #f2f4f7;
   min-height: 100vh;
-  width: 100%;
+  min-width: 1024px;
   font-family: "Poppins",sans-serif;
 }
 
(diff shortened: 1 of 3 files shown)


src/features/rewards/welcomePage/__snapshots__/spec.tsx.snap
```diff
@@ -15,7 +15,7 @@ exports[`WelcomePage tests basic tests matches the snapshot 1`] = `
 .c0 {
   background: #f2f4f7;
   min-height: 100vh;
-  width: 100%;
+  min-width: 1024px;
```
ryanml: Note: This enforces the same behavior on the desktop welcome page, which seems desirable as we use the same SettingsPage component. This will **not** however, do so for the mobile welcome page, as it does not render inside the SettingsPage component.

NejcZdovc:
@jenn-rhim is this what you wanted, or would you like to stack them?

PR #239, opened 2018-10-24
additional changes in the mobile view

jenn-rhim:
to improve legibility


ryanml:
@jenn-rhim - can you commit the updated snapshot files for these edited components?

Also, package-lock.json does not need to be changed, if you could remove it from these changes thanks.

PR #241, opened 2018-10-25
Adds new 2 state tab component

ryanml:
Pictured in storybook:

screen shot 2018-10-24 at 7 45 26 pm

### DIFF
--- a/src/features/rewards/tab/index.tsx
+++ b/src/features/rewards/tab/index.tsx
@@ -1,0 +1,67 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this file,
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react'
+import {
+  RewardsTabWrapper,
+  StyledSlider,
+  StyledBullet,
+  StyledText,
+  StyledTab,
+  StyledSwitch
+} from './style'
+
+export interface Props {
+  tabTitles?: string[]
+  testId?: string
+  tabIndexSelected?: number
+  onChange?: (event: React.MouseEvent<HTMLDivElement>) => void
+}
+
+export default class Tab extends React.PureComponent<Props, {}> {
+  static defaultProps = {
+    tabIndexSelected: 0
+  }
+
+  getTabs = (tabTitles: string[]) => {
+    const tabs: React.ReactNode[] = tabTitles.map((title: string, i: number) => {
+      return (
+        <StyledTab
+          left={i === 0}
+          key={`tab-${i}`}
+        >
+          <StyledText selected={i === this.props.tabIndexSelected}>
+            {title}
+          </StyledText>
+        </StyledTab>
+      )
+    })
+    return tabs
+  }
+
+  render () {
+    const {
+      testId,
+      tabTitles,
+      onChange,
+      tabIndexSelected
+    } = this.props
+
+    if (!tabTitles || tabTitles.length !== 2) {
+      console.warn('Rewards Tab currently supports 2 tab titles')
+      return null
+    }
+
+    return (
+      <RewardsTabWrapper>
+        <StyledSwitch onClick={onChange}>
+          <StyledSlider data-test-id={testId}>
+            {this.getTabs(tabTitles)}
+          </StyledSlider>
+          <StyledBullet tabIndexSelected={tabIndexSelected} />
+        </StyledSwitch>
+      </RewardsTabWrapper>
+    )
+  }
+}
--- a/src/features/rewards/tab/style.ts
+++ b/src/features/rewards/tab/style.ts
@@ -1,0 +1,63 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License. v. 2.0. If a copy of the MPL was not distributed with this file.
+ * You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import styled from 'styled-components'
+import { Props } from './index'
+
+interface StyleProps {
+  left?: boolean
+  selected?: boolean
+}
+
+export const RewardsTabWrapper = styled<{}, 'div'>('div')`
+  display: flex;
+  font-family: Poppins,sans-serif;
+`
+
+export const StyledSwitch = styled<{}, 'div'>('div')`
+  position: relative;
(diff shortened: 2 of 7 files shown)


src/features/rewards/tab/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+    const {
+      id,
+      tabOne,
+      tabTwo,
+      onSwitch,
+      tabOneChecked
+    } = this.props
+
+    return (
+      <StyledWrapper id={id}>
```
petemill: IMO let's start giving 'StyledWrapper' a more unique name - when inspecting the code almost everything is `StyledWrapper` class, and I'd like to see something indicating what component it is. In this case, we could call the outer element the `RewardsTabWrapper`

src/features/rewards/tab/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  StyledWrapper,
+  StyledSlider,
+  StyledBullet,
+  StyledTextOne,
+  StyledTextTwo,
+  StyledSwitch
+} from './style'
+
+export interface Props {
+  id?: string
```
petemill: Can we use `testId` and `data-test-id` attribute or did I miss a change in process for that?

src/features/rewards/tab/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  tabOneChecked?: boolean
+  tabOne?: string
+  tabTwo?: string
+  onSwitch?: (event: React.MouseEvent<HTMLDivElement>) => void
```
petemill: Some suggestions for slightly better props naming:
- `onChange` as the event handler, and provide the index as the parameter instead of a MouseEvent
- `tabTitles` as a string array (throw an error if more than two are provided for now)
- `tabIndexSelected`

src/features/rewards/tab/index.tsx
ryanml: done

src/features/rewards/tab/index.tsx
ryanml: done

src/features/rewards/tab/index.tsx
ryanml: done

src/features/rewards/tab/index.tsx
```diff
@@ -0,0 +1,67 @@
...
+      tabIndexSelected
+    } = this.props
+
+    if (!tabTitles || tabTitles.length !== 2) {
+      console.warn('Rewards Tab currently supports 2 tab titles')
+      return null
+    }
+
+    return (
+      <RewardsTabWrapper id={id}>
```
petemill: Do we need this extra wrapper element? Doesn't seem to do much.

src/features/rewards/tab/style.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const RewardsTabWrapper = styled<{}, 'div'>('div')`
+  display: flex;
+  font-family: Poppins,sans-serif;
+`
+
+export const StyledSwitch = styled<{}, 'div'>('div')`
+  position: relative;
+  display: block;
+  width: 284px;
```
petemill: Looks like we're using fixed widths for:
- The component overall
- Each item

This seems problematic for items that have a longer string. Whilst we may only have one usage of this and it only supports 2 Tabs so we can ignore a swathe of detail considerations, we of course support a high number of languages. Can we predict that the word for 'Backup' and 'Restore' will fit in to 135px at 16px Poppins? The first language I looked up (Finnish) had the word for 'Backup' as 'Varmuuskopioida':
image

@jenn-rhim what is the intention here:
- each item takes up the space it needs, and the 'pill' (called 'bullet' here) grows and shrinks as it moves between selected items?
- each item is the same width, taking the largest item's required width
- each item is a fixed size we define keeping 🤞 that we cater to all language string lengths

This is a happy reminder that thoughts and detail like this is why it's important to put effort in to creating a shared spec for a complex custom new thing like this and want to maintain quality @karenkliu @bradleyrichter @jenn-rhim 😃

petemill: One option for this is to:
- Set a max-width on the overall component (or let it be 100% width of its parent)
- Allow each item to take up the width it needs, but break the word on to the next line if it's too long, even if it's 1-word. So we will have varying width items.
- All items take the same height, so if one item takes up two lines, all the other items take up 2 lines too.

There are many options, but that is 1, and maybe it is not the easiest to implement.

jenn-rhim: @pete thanks for checking a translated string. Until we run into a situation that it becomes a problem, each tab should be given the same width, considering the longest string item, which will provide the visual balance. I don’t expect the word to break into 2 lines if it’s a single word, which in this case it is. Also don’t expect to use multi-word categories for tab headings - tab control probably is not the right choice of the design for that case. There are also other ways to shrink the text. But let’s not worry about it for now. 

Consider using the font size 14 instead of 16. I’d actually prefer the style that leaves more white space within the tab element than filling it.

src/features/rewards/tab/index.tsx
jenn-rhim: We should be able to support all of the above mentioned @pete. Let me know why it can’t.

jenn-rhim: We should be able to support all of the above mentioned @pete. Let me know why it can’t.

jenn-rhim: Fyi, I don’t want to speak for @ryanml but he may not have taken those into account because he was implementing it for the android vs.

src/features/rewards/tab/style.ts
ryanml: 1. The tab component now takes up 100% of the parent width, and divides tab space equally.
2. Text should now truncate at a reasonable limit based on the tab size
3. Font reduced to 14px

Here is the element at a few widths:

500px
screen shot 2018-10-29 at 6 12 45 pm

350px
screen shot 2018-10-29 at 6 12 55 pm

200px
screen shot 2018-10-29 at 6 13 08 pm

cc: @petemill @NejcZdovc @jenn-rhim

src/features/rewards/tab/index.tsx
ryanml: @jenn-rhim what are the expected active and hover states for this component?

PR #243, opened 2018-10-25
style updates

jenn-rhim:


NejcZdovc:
@jenn-rhim please update snapshots

PR #245, opened 2018-10-26
fix check for ontouchstart

cezaraugusto:


cezaraugusto:
`ontouchstart` is a global event so any DOM element will return it. didn't try it with window before but more elegant so there we go

PR #246, opened 2018-10-27
WIP - Icon updates

rossmoody:
Generally speaking these were meant to just be icon updates for sizing but it became a little more. Basically our icons were 2px too small within the bounding box at 32px (the size they are being made originally). This was causing them bounding box to need to be too big in many cases to get the icon at the size it needed to be to look visually balanced. Across the board the icons were updated to be 26px tall at the largest width or height. Some additional updates were also made including:

- styling for the storybook icons page, these are loose and iterative
- order of the components on the index page 
- the naming convention for icons being pulled in full color instances. this was updated to reflect "fc-logo-name of brand" 

**Before**

image

**After**

image


rossmoody:
The carat changes on this will need gone over with a fine tooth comb as they are used in lots of places and out of the whole batch experienced the largest size increase.

Issue #252, opened 2018-10-31
Create component for imported pinned sites notification


ryanml:
This component will appear below the Box component (tips variant) when an `onReview` function prop is present. That function prop will be passed in to the new component.

Pictured here:

<img width="628" alt="screen shot 2018-10-30 at 10 41 25 pm" src="https://user-images.githubusercontent.com/8732757/47768435-5b039780-dc95-11e8-9780-1a0b11a79f4e.png">

PR #253, opened 2018-10-31
Ads tipsMigrationAlert component

ryanml:
Fixes: #252 
Related: https://github.com/brave/brave-browser/issues/1910

screen shot 2018-10-31 at 4 14 59 am

Edit, updated design:

screen shot 2018-11-01 at 11 42 26 pm


src/features/rewards/tipsMigrationAlert/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  StyledMessage,
+  StyledMonthlyTips,
+  StyledPrompt,
+  StyledReviewWrapper,
+  StyledReviewList
+} from './style'
+import { AlertCircleIcon } from '../../../components/icons'
+
+export interface Props {
+  id?: string
```
petemill: can we use data-test-id instead of id?

src/features/rewards/box/index.tsx
```diff
@@ -32,6 +33,7 @@ export interface Props {
   description?: string
   toggle?: boolean
   checked?: boolean
+  onReview?: () => void
```
petemill: As far as I can see, the Box component is pretty generic. So 'onReview' is a bit confusing because not every box will be reviewable, and the box itself isn't for review. Instead, could the box component accept props about the alert it should display, if any. For example:
`alertData: { onReview, alertTitle, alertDescription, alertAction }`

Actually in this case, since it's so specific and we have a custom component just for this alert, why not have the alert component passed through as a prop?

`attachedAlert: React.ReactNode`

src/features/rewards/tipsMigrationAlert/style.ts
```diff
@@ -0,0 +1,63 @@
...
+*/
+
+import styled from 'styled-components'
+
+export const StyledWrapper = styled<{}, 'div'>('div')`
+  width: 100%;
+  display: block;
+  background: #E9F0FF;
+  padding: 5px 30px;
+  border-bottom-left-radius: 6px;
```
petemill: Since this component ends up inside the box, can it just have itself be clipped by its container? I.e. keep the border radius on the Box component, and have this alert be the last child of the Box?

NejcZdovc: @ryanml you could just add border radius to new wrapper that you created for box, so that you can remove this one https://github.com/brave/brave-ui/pull/253/files#diff-6bd2b80c49423c16c202f58fbb1010b3R34

NejcZdovc:
@ryanml there  is new design for it, can you please update it? https://github.com/brave/brave-browser/issues/1910#issuecomment-435238110

src/features/rewards/box/index.tsx
NejcZdovc: sure I like `attachedAlert: React.ReactNode`

src/features/rewards/tipsMigrationAlert/index.tsx
ryanml: done

src/features/rewards/box/index.tsx
ryanml: done 👍

src/features/rewards/tipsMigrationAlert/style.ts
ryanml: done

NejcZdovc:
@ryanml please use flex instead of inline-block, so that this scenario is handled as well

image

src/features/rewards/tipsMigrationAlert/style.ts
```diff
@@ -0,0 +1,59 @@
...
+  border-bottom-right-radius: 6px;
+`
+
+export const StyledAlertIcon = styled<{}, 'div'>('div')`
+  width: 57px;
+  height: 57px;
+  color: #15A4FA;
+  display: inline-block;
```
NejcZdovc: you don't need inline-block anymore now that you have flex, please remove it in all elements

src/features/rewards/tipsMigrationAlert/style.ts
```diff
@@ -0,0 +1,56 @@
...
+export const StyledMessage = styled<{}, 'span'>('span')`
+  color: #000;
+  font-weight: 400;
+  margin-right: 3px;
+`
+
+export const StyledMonthlyTips = styled<{}, 'span'>('span')`
+  color: #696FDC;
+  font-weight: 400;
+  display: inline-block;
```
ryanml: we need this one

src/features/rewards/tipsMigrationAlert/style.ts
ryanml: @NejcZdovc done

Issue #254, opened 2018-11-02
Create component for mobile site banner


ryanml:
Note: This will not require a separate component, this should be more than doable via CSS

Concept pictured here:

![tipping-banner-alt](https://user-images.githubusercontent.com/8732757/47895669-e52f3580-de26-11e8-82a2-74ff8da72f31.png)

PR #255, opened 2018-11-02
Mobile siteBanner

ryanml:
Fixes: #254 

Note: A new component was not created for this, as layout did not really need to change, just styles.

Now Link: https://brave-ui-awdllyyigp.now.sh
(Use the mobile inspector and view at **Feature Components > Rewards > Concepts > Mobile > Site Banner**)

In Action:

screen shot 2018-11-02 at 11 52 04 pm

### DIFF
--- a/src/features/rewards/amount/__snapshots__/spec.tsx.snap
+++ b/src/features/rewards/amount/__snapshots__/spec.tsx.snap
@@ -2,7 +2,6 @@
 
 exports[`Amount tests basic tests matches the snapshot 1`] = `
 .c0 {
-  margin-bottom: 8px;
   -webkit-user-select: none;
   -moz-user-select: none;
   -ms-user-select: none;
@@ -20,6 +19,7 @@ exports[`Amount tests basic tests matches the snapshot 1`] = `
   -webkit-box-align: center;
   -ms-flex-align: center;
   align-items: center;
+  margin: 0 0 8px 0;
 }
 
 .c1 {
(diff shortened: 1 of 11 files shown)


jenn-rhim:
@ryanml looks great! Maybe the site initial can be smaller. Other than that, I think it’s ready for a spin. Btw, looks like the string update has not made it in here yet— still seeing ‘donation’ which should be ‘Tip’.

ryanml:
@jenn-rhim : Site initial has been updated to be smaller (new link is: https://brave-ui-awdllyyigp.now.sh/, preview in PR updated as well)

https://github.com/brave/brave-ui/issues/256

I just made this issue regarding the string update, as we'd be removing/modifying translated strings, I'd rather this be accomplished in a separate issue. Thanks!

cezaraugusto:
@ryanml please rebase

ryanml:
@cezaraugusto rebased

ryanml:
Here is this PR overlaying master: (I think the overlay above may be mismatched)

As you can see, the site banner portion on desktop should be unmodified. The only modifications I made explicitly on desktop as well were to the "Send my Donation" button, as it was crooked on desktop. I think we should keep this modification.

I see the spacing is a bit different on this PR for donation amount buttons, that I can fix back. wdyt about keeping the modifications to the send button? cc: @jenn-rhim @NejcZdovc 

screen shot 2018-11-08 at 11 15 39 am

Issue #262, opened 2018-11-05
Add info modal to tipsMigrationAlert


ryanml:
Related to: https://github.com/brave/brave-browser/issues/1910#issuecomment-435238110

Pictured: https://user-images.githubusercontent.com/4369856/47888172-1d6f4d80-de00-11e8-9434-8ea336d5d919.png

PR #263, opened 2018-11-05
Adds pinned sites info overlay

ryanml:
Fixes: #262


src/features/rewards/tipsMigrationAlert/index.tsx
```diff
@@ -3,28 +3,112 @@
...
+  StyledListItem,
+  StyledTipsIcon,
+  StyledButton
 } from './style'
+import { Modal } from '../../../components'
 import { AlertCircleIcon } from '../../../components/icons'
+import Button from '../../../components/buttonsIndicators/button'
+
+// Assets
+const sendTipsImage = require('./assets/sendTips')
```
NejcZdovc: please use import here and export default in image file, so that is inlined with @petemill work

PR #268, opened 2018-11-06
Adding RestoreSites component

ryanml:
Related: https://github.com/brave/brave-browser/issues/1883

Since code would have to be duplicated to include this in `TableContribute`, I refactored the restore sites prompt in to its own component.

Pictured here on the `TableContribute` component under `Box`:

screen shot 2018-11-05 at 5 31 19 pm

Note ^^ as this prompt is hidden on the outer view when "Show All" is present, you won't see it like this in the storybook. You may view the component in the contribute modal by clicking "Show All" in the event of 5 or few sites, this will be visible on the outer view.

### DIFF
--- a/src/features/rewards/restoreSites/index.tsx
+++ b/src/features/rewards/restoreSites/index.tsx
@@ -1,0 +1,39 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this file,
+* You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+import * as React from 'react'
+import { getLocale } from '../../../helpers'
+import {
+  StyledExcludedText,
+  StyledRestore
+} from './style'
+
+export interface Props {
+  numExcludedSites: number
+  onRestore?: () => void
+}
+
+export default class RestoreSites extends React.PureComponent<Props, {}> {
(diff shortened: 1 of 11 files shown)


src/features/rewards/restoreSites/style.ts
```diff
@@ -0,0 +1,30 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this file,
+* You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+import styled from 'styled-components'
+
+export const StyledWrapper = styled<{}, 'div'>('div')`
+  display: flex;
```
NejcZdovc: Is flex needed here?

src/features/rewards/restoreSites/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+  getExclusionText (numSites: number) {
+    return `${getLocale('excludedSitesText')} ${numSites}`
+  }
+
+  render () {
+    const { numExcludedSites, onRestore } = this.props
+
+    return (
+      <StyledWrapper>
```
NejcZdovc: Do we need wrapper?

ryanml: removed

PR #270, opened 2018-11-08
Changes Donations -> Tips

### LINKED ISSUE #256: Strings: References to "Donation" should be changed to "Tip"

ryanml:
Fixes #256 , our storybook should be caught up with the current browser


NejcZdovc:
@ryanml for now let's just change text to be up to date, don't want to update name of classes and variables yet. We have the same problem in core / native ledger. Would like to stabilise code first and then do this kind of refactors. Thank you

PR #274, opened 2018-11-13
Implements cross-platform Backup/Recovery modal

ryanml:
Live storybook link: https://brave-ui-ubfmotahgl.now.sh

Preview:
(Slight differences in design vs shared components)
screen shot 2018-11-12 at 10 47 41 pm


src/components/formControls/textarea/index.tsx
```diff
@@ -12,12 +12,13 @@ export interface Props {
   readOnly?: boolean
   placeholder?: string
   disabled?: boolean
+  fieldError?: boolean
```
ryanml: Per designs, we need to be able to highlight the text area error red in case of a problem

cc: @jenn-rhim

src/features/rewards/modalBackupRestore/index.tsx
```diff
@@ -77,145 +84,200 @@ export default class ModalBackupRestore extends React.PureComponent<Props, State
...
   setRecoveryKey = (event: React.ChangeEvent<HTMLTextAreaElement>) => {
     this.setState({
+      errorShown: false,
       recoveryKey: event.target.value
     })
   }
 
   onRestore = (key?: string) => {
     key = typeof key === 'string' ? key : this.state.recoveryKey
-    this.setState({
```
ryanml: Per design, we want to retain the recovery key in case of a problem

cc: @jenn-rhim

src/features/rewards/modalBackupRestore/index.tsx
```diff
@@ -33,10 +37,12 @@ export interface Props {
   error?: React.ReactNode
   id?: string
   testId?: string
+  funds?: string
```
ryanml: We need to be able to produce a special warning for restore if the user has funds in their wallet.

What should be fed in this prop from assembly is a formatted string of the users BAT balance. Ex: '55 BAT'

PR #278, opened 2018-11-16
Rewards Design working session updates

ryanml:
This PR is the culmination of the design updates/fixes that were made today during a working session with @jenn-rhim and myself.

Changes include:

Adding **rewardsButton**
- This component was added to achieve the desired design for the button in the wallet panel as well as       the welcome page opt-in buttons. It was made to avoid modifications to the shared button component. This was a higher priority of the updates.

Styling/locale string updates and improvements made to components and concepts of the Wallet panel, Rewards Welcome page and Rewards settings.
- A lot of these focused on font weight and font family consistency.

Updating various rewards icons


src/features/rewards/walletWrapper/index.tsx
```diff
@@ -45,12 +45,13 @@ import {
...
   SettingsAdvancedIcon,
   UpholdColorIcon,
-  UpholdSystemIcon,
-  LoveIcon,
-  GiftIcon,
-  MegaphoneIcon
+  UpholdSystemIcon
 } from '../../../components/icons'
 
+import giftIcon from './assets/gift'
```
NejcZdovc: how come that we switched from icon to asset?

rossmoody: it looks like it's cause the icon is full color now. could possibly move this to a shared full color icon component but i don't think it's causing harm and could be considered feature specific.

petemill: I would actually like to have these assets go in as separate actual SVG files, for performance reasons. I’ll follow up with how that’ll be possible with our new Webpack config once that’s done in brave-core. We could also get the same gains by using import() syntax. 
Inline SVG should only really be used for small amounts of data when we want to affect change through css.

ryanml: import() syntax will be used in this PR for added svg's once https://github.com/brave/brave-ui/pull/282 has been merged. Thanks!

PR #285, opened 2018-11-21
Upgrades and import blahUrl from 'blah.svg' syntax for file-loader

petemill:
Upgrading storybook was yak shaving in order to get latest babel which was yak shaving in order to make the build output and ts-jest happy with using es6 modules output from typescript, instead of nodejs `require` syntax.

We were just trying to get `import blahUrl from `blah.svg` syntax working, which it does in the end!

I had to convert all the `require('blah.svg')` uses since the file mock either has to use `module.exports` or `export default`, it can't do both.


petemill:
@ryanml oops yes I didn't upgrade the integration test's typescript, jest and webpack config

Issue #298, opened 2018-11-30
Ensure that amount related functions expect string type


ryanml:
This overhaul was originally done here:

https://github.com/brave/brave-ui/pull/194

Integration tests are currently failing:
<img width="790" alt="screen shot 2018-11-30 at 12 23 38 am" src="https://user-images.githubusercontent.com/8732757/49274705-700b4c00-f436-11e8-8897-83f8f5cd1569.png">

Amount types should be changed to string, integration tests should pass, and UI should work with amount as expected.

PR #299, opened 2018-11-30
Uses string type for amount/tokens

ryanml:
Fixes: https://github.com/brave/brave-ui/issues/298

This is what involved in fixing the integration test error in the issue. It seems that there was further inconsistency in amount typing even after the work done https://github.com/brave/brave-ui/pull/194

Everything works in the UI as far as I have tested, (Balance calculation works, and expected error will come up if you select a token amount that exceeds your wallet balance

### DIFF
--- a/src/features/rewards/amount/index.tsx
+++ b/src/features/rewards/amount/index.tsx
@@ -10,7 +10,7 @@ import { BatColorIcon } from '../../../components/icons'
 export interface Props {
   amount: string
   converted: string
-  onSelect: (amount: number) => void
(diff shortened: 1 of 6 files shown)


ryanml:
@NejcZdovc @petemill and I tested this PR as is in brave-core, and all built fine with no runtime errors that could be soon.

Amount in the donate banner on core is a `number` type so these changes work nicely.

I just wanted to confirm this integration worked, we should be fine to merge this, publish to npm, then update version in core.

PR #301, opened 2018-11-30
Add support for insufficientFunds notification

emerick:
This is associated with brave/brave-browser#1479.

Adds support for "insufficient funds" notifications in Rewards, which is sent when the current balance is insufficient to make new contributions.


emerick:
@ryanml Thank you!

PR #304, opened 2018-12-05
Adding disabledPanel component

ryanml:
Related: https://github.com/brave/brave-browser/issues/2214

Now Link: https://brave-ui-bdjtqnnmba.now.sh

Storybook:

screen shot 2018-12-04 at 8 24 22 pm

Can be viewed under: Feature Components > Rewards > Concept > Desktop > Disabled Panel


src/features/rewards/disabledPanel/style.ts
```diff
@@ -0,0 +1,55 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this file,
+* You can obtain one at http://mozilla.org/MPL/2.0/.
+*/
+
+import styled from 'styled-components'
+
+export const StyledWrapper = styled<{}, 'div'>('div')`
+  width: 330px;
```
NejcZdovc: I think width should be 100% and panel should define it

src/features/rewards/disabledPanel/style.ts
```diff
@@ -0,0 +1,55 @@
...
+  padding: 30px 25px 25px;
+  border-radius: 8px;
```
NejcZdovc: we need to remove it as we can't have radius in panel

src/features/rewards/disabledPanel/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  StyledText,
+  StyledLink,
+  StyledTitleWrapper,
+  StyledTextWrapper
+} from './style'
+
+import { getLocale } from '../../../helpers'
+
+export interface Props {
+  id?: string
```
NejcZdovc: I think we are using test-id or something now? could be mistaken

src/features/rewards/disabledPanel/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <StyledWrapper id={id}>
+        <StyledTitleWrapper>
+          <StyledTitle>
+            {getLocale('disabledPanelTitle')}
+          </StyledTitle>
+          <StyledOff>
+            {getLocale('disabledPanelOff')}
+          </StyledOff>
+        </StyledTitleWrapper>
+        <StyledTextWrapper>
```
NejcZdovc: Do we need wrapper and then text? Can we just have text?

src/features/rewards/disabledPanel/index.tsx
ryanml: I don't think we are explicitly supposed to be using something but yeah this should probably be testId instead of Id

src/features/rewards/disabledPanel/index.tsx
ryanml: I have a few rules on on `StyledTextWrapper` that are being used. 

In the case where I have text containing two components, in this cased `StyledText` and `StyledLink`, I prefer to have a parent container

src/features/rewards/disabledPanel/index.tsx
ryanml: added

src/features/rewards/disabledPanel/style.ts
ryanml: done

src/features/rewards/disabledPanel/style.ts
ryanml: yep, done

src/features/rewards/disabledPanel/index.tsx
NejcZdovc: but isn't `StyledLink` inside `StyledText`

NejcZdovc: so `StyledTextWrapper` has only one direct child component

ryanml: Lol yes it is, I misread my own code as it's a pattern I frequently use, I can update

PR #305, opened 2018-12-05
Fixing rewards toggle for long translations

ryanml:
Related: https://github.com/brave/browser-android-tabs/issues/914

Fix at small (320px) viewport: (Text wrapping was agreed upon)
screen shot 2018-12-10 at 4 47 41 pm

Fix at larger, but still slim 375px viewport:
screen shot 2018-12-10 at 4 47 55 pm

### DIFF
--- a/src/features/rewards/mobile/mainToggleMobile/style.ts
+++ b/src/features/rewards/mobile/mainToggleMobile/style.ts
@@ -30,17 +30,10 @@ export const StyledLeft = styled<{}, 'div'>('div')`
 `
 
 export const StyledRight = styled<{}, 'div'>('div')`
-  padding: 0px 15px;
-  display: flex;
   height: 66px;
-  align-items: center;
-  margin-top: -3px;
-
-  @media (max-width: 355px) {
-    width: 55px;
-    position: relative;
-    right: 45px;
-  }
+  position: fixed;
+  right: 15px;
+  top: 18px;
 `
 
 export const StyledTitle = styled<{}, 'div'>('div')`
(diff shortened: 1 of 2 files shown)


src/features/rewards/mobile/mainToggleMobile/style.ts
```diff
@@ -30,26 +30,27 @@ export const StyledLeft = styled<{}, 'div'>('div')`
...
   height: 66px;
-  align-items: center;
-  margin-top: -3px;
-
-  @media (max-width: 355px) {
-    width: 55px;
-    position: relative;
-    right: 45px;
-  }
+  position: fixed;
```
petemill: This seems weird to have the checkbox fixed position on a header that is already fixed position. If the aim is to vertically center the checkbox then, since the parent is flex, it can be achieved via:
```css
align-self: center;
margin-right: 15px;
```
Although preferably the parent would specify a padding instead of the element individually specifying its right margin.

src/features/rewards/mobile/mainToggleMobile/style.ts
```diff
@@ -30,26 +30,27 @@ export const StyledLeft = styled<{}, 'div'>('div')`
...
-  line-height: 44px;
   margin-bottom: -23px;
+  line-height: 44px;
+
+  @media (max-width: 375px) {
+    max-width: 250px;
+    word-wrap: break-word;
+    display: block;
+    line-height: 20px;
+    margin-top: -12px;
```
petemill: hmm, do we need all this complicated negative margins and line-height? If the aim is to have the text be vertically centered whether it's displayed on 1 line or 2, can't we have the parent be 100% height and just vertically center the text via flexbox, and then allowing wrapping at the specific width as you have here?

src/features/rewards/mobile/mainToggleMobile/style.ts
```diff
@@ -61,6 +62,10 @@ export const StyledTM = styled<{}, 'span'>('span')`
   position: relative;
   top: -13px;
   vertical-align: text-top;
+
+  @media (max-width: 375px) {
```
petemill: Instead of using negative margin, desktop uses flex alignment to get aligned to the top of the box around the text - could that simplify this?

PR #308, opened 2018-12-06
make sync images exportable

cezaraugusto:
addresses partial feedback from https://github.com/brave/brave-core/pull/1019


src/features/sync/images/index.ts
```diff
@@ -0,0 +1,46 @@
...
+
+import styled from '../../../theme'
+
+import StartIconImage from './start_icon.svg'
+import DefaultIconImage from './default_icon.svg'
+import AddIconImage from './add_icon.svg'
+import RemoveIconImage from './remove_icon.svg'
+import DesktopImage from './desktop_icon.svg'
+import MobilePicture from './mobile_icon.svg'
+import MobilePictureImage from './mobile_picture.png'
```
petemill: nit: there'd be less confusion in this module of variable names if the urls ended in Url and the actual image elements ended in Image (or Icon or Picture I guess). But at the moment they are mixed - some urls end in Picture and some components end in Picture, etc.

cezaraugusto: k added the `url` suffix and normalized them to use `image` instead of icon/picture

PR #311, opened 2018-12-06
Use new static deploy config for now instead of docker

### LINKED ISSUE #310: now deployments are unreliable
Takes a long time, sometimes doesn't result in a working Url

petemill:
Fix #310

We used docker because previously now did not support custom commands or paths. Now it does, so we can use standard npm install, npm command, and static site.

This does not affect anything else and does not need an `npm publish` after merge. It just makes `now -p` more reliable.


petemill:
You need to update your local ‘now’. I guess we could put it as a dependency and then put the actual ‘now’ command under ‘npm run build’

PR #312, opened 2018-12-07
Add support for backupWallet notification

emerick:
This is associated with brave/brave-browser#1164.

Adds support for "backup wallet" notifications in Rewards, which is sent when the user needs to backup their wallet.

### DIFF
--- a/src/features/rewards/walletWrapper/index.tsx
+++ b/src/features/rewards/walletWrapper/index.tsx
@@ -69,7 +69,7 @@ export interface ActionWallet {
   action: () => void
 }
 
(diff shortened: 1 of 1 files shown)


src/features/rewards/walletWrapper/index.tsx
```diff
@@ -166,6 +166,9 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
       case 'ads':
         icon = megaphoneIconUrl
         break
+      case 'backupWallet':
```
NejcZdovc: can we merge it with `ads` case above?

emerick: Makes sense, done!

PR #313, opened 2018-12-11
Adding routing for mobile rewards settings views

ryanml:
Related: https://github.com/brave/browser-android-tabs/issues/905

Adds support for the following (generated based off of `type` prop)

`/`
`/ads`
`ads-settings`
`/contribute`
`contribute-settings`
`/donate`
`donate-settings`


petemill:
I pushed an example app (and a rebase on master) to view the Rewards Settings (mobile) page standalone in order to test the url handling:

```
cd examples/rewards-mobile-settings
npm i
npm start
```

It handles changing URL correctly, but does not handling back button correctly. As far as I can tell, there's no `popstate` handler, meaning the code won't be informed when there's a new URL to respond to. I'll see if I can quickly add that in to the page @ryanml

PR #317, opened 2018-12-11
Create a pull request template for integration process

petemill:
Checklist for npm publish or branch integration

Obviously this PR does not require integration

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,22 @@
+## Changes
+
+## Test plan
+
+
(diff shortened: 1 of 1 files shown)


petemill:
@ryanml Interesting point. I put the questions there as indicators for the person merging or integrating to work out what kind of release would need to be done. I'm not sure that a dependency change requires any further action other than the review? Plus, a change in package.json is a clearer indicator compared to anything that exists to automatically answer the questions about the types of changes contained.

PR #344, opened 2019-01-10
Add support for the default new tab page

cezaraugusto:
test plan:

should match brave-core current (nightly) new tab page.

heads-up I replaced react-dnd with react-beautiful-dnd in this change. this would be a follow-up task in brave-core and can be skipped in this review if you prefer.


src/features/newTab/default/topSites/index.ts
```diff
@@ -97,7 +96,6 @@ export const Tile = styled<TileProps, 'div'>('div')`
 `
 
 export const TileFavicon = styled<{}, 'img'>('img')`
-  box-sizing: border-box;
   display: block;
-  height: 64px;
+  height: 40px;
 `
```
rossmoody: @cezaraugusto just wanted to bring attention to the size update for the favicons here. i noticed some updates regarding favicon sizing in the Core PR and wasn't sure if this would mess it up.

cezaraugusto: ya I updated brave-core to fetch 64px favicons as well

PR #347, opened 2019-01-14
Fix clock component path

rossmoody:
My storybook run command was failing, this fixed it for me.

### DIFF
--- a/src/features/newTab/default/clock/index.tsx
+++ b/src/features/newTab/default/clock/index.tsx
@@ -66,3 +66,5 @@ export class Clock extends React.PureComponent<{}, ClockState> {
     )
   }
 }
+
+export default Clock
--- a/src/features/newTab/default/clock/index.tsx
+++ b/src/features/newTab/default/clock/index.tsx
@@ -66,5 +66,3 @@ export class Clock extends React.PureComponent<{}, ClockState> {
     )
   }
 }
-
-export default Clock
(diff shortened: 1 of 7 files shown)


stories/old/views.story.tsx
```diff
@@ -26,7 +26,7 @@ import * as React from 'react'
 // TBD import Clock from '../components/clock'
 import Page from '../../src/old/page'
 import { DataBlock, DataItem } from '../../src/old/dataBlock'
-import Clock from '../../src/old/clock'
+import Clock from '../../src/features/newTab/default/clock/index'
```
cezaraugusto: please change it to `import { Clock } from '../../src/features/newTab/default/clock`

stories/features/newTab/default/index.tsx
```diff
@@ -5,7 +5,7 @@
 import * as React from 'react'
 
 // Feature-specific components
-import Clock from '../../../../src/old/clock'
+import Clock from '../../../../src/features/newTab/default/clock/index'
```
cezaraugusto: please change it to `import { Clock } from '../../../../src/features/newTab/default/clock'`

src/features/newTab/default/clock/index.tsx
```diff
@@ -66,3 +66,5 @@ export class Clock extends React.PureComponent<{}, ClockState> {
     )
   }
 }
+
+export default Clock
```
cezaraugusto: could you remove this line? we are exporting this component but not as default. see the other comments, I think we can get ride of it if all changes are applied

PR #355, opened 2019-01-25
Close Icon Refresh

rossmoody:
Fixes https://github.com/brave/brave-browser/issues/3111 where the updates are visually outlined.


mandar-brave:
@rossmoody these look good to go!
Thanks.

PR #365, opened 2019-02-05
Opt In Button 

rossmoody:
Addresses: https://github.com/brave/brave-browser/issues/3218

Since there are upcoming button component styling consolidation initiatives, I think adding this prop to the base button component to allow for white buttons instead of always defaulting to `subtle` is helpful in the interim. Regardless, white buttons will be a needed styling so this style is forthcoming anyway.

## Before
screen shot 2019-02-04 at 4 39 58 pm

## After
image

### DIFF
--- a/src/components/buttonsIndicators/button/index.tsx
+++ b/src/components/buttonsIndicators/button/index.tsx
@@ -26,7 +26,7 @@ export interface Props {
 }
 
 export type Level = 'primary' | 'secondary' | 'tertiary'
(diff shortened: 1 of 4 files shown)


rossmoody:
Got me thinking about this, would it make more sense to refer to is `inverse`, perhaps? The major need for this right now is we don't have a way to surface light buttons on dark backgrounds that aren't `subtle`, right?

rossmoody:
Gonna close this one down and just wait for the proper button definitions down the line.

PR #382, opened 2019-02-20
Shields v2.1

cezaraugusto:
shield

address https://github.com/brave/brave-browser/issues/2565
address https://github.com/brave/brave-browser/issues/1894
address https://github.com/brave/brave-browser/issues/1892

## Changes

Mostly refactor everything under `shields/`

## Test plan

**Note:** There was a change in the `Toggle` element that is also shared with the private new tab. I'll handle this when porting to brave-core.

**Note 2:** Design feedback and review was made by @karenkliu and should be good to go unless her or someone else from the design team says otherwise.

1. Ensure you can navigate through shields using the tab key and that the space key trigger the action that `onClick` would trigger.

- https://brave-ui-959w0ii2u.now.sh

## Integration
  
  - [x] Are there non backwards-compatible changes required for brave-core? **Do not merge until brave-core PR is approvable.** Link to brave-core PR:
  - [x] Will you create brave-core PR to update to this commit after it is merged?
  - [x] Wants uplift to brave-core feature branch?


stories/features/shields/components/controls/scriptsControl.tsx
```diff
@@ -0,0 +1,108 @@
...
+    this.state = {
+      scriptsBlockedOpen: false,
+      scriptsBlockedEnabled: true
+    }
+  }
+
+  get tabIndex () {
+    const { isBlockedListOpen } = this.props
+    return isBlockedListOpen ? -1 : 0
+  }
```
rossmoody: From my testing, I think we might want to forego the tabIndex functionality for now. It's awesome we have it and the focus box looks great, but I think we don't have the best parts of tabIndex wired up yet (like pressing Enter on keyboard to submit or change the toggle and then remove focus) so we are surfacing the orange box focus ring after any interactivity with toggles but then it doesn't go away and makes the component look like it's still active even though it's not. I'd vote for having this be inherited functionality when we consolidate Brave-UI components.

src/features/shields/structure.ts
```diff
@@ -0,0 +1,318 @@
...
+
+interface ShieldsHeaderProps {
+  status: 'enabled' | 'disabled'
+}
+
+export const ShieldsHeader = styled<ShieldsHeaderProps, 'header'>('header')`
+  box-sizing: border-box;
+  background-image: url(${Background});
+  background-repeat: no-repeat;
+  background-position: ${p => p.status === 'enabled' ? '0 -15px' : '0 0'};
```
rossmoody: 100% personal preference, but thought this element was riding a little high:
`background-position: ${p => p.status === 'enabled' ? '0 -5px' : '0 0'};`

rossmoody:
Love the interactivity polish here. So nice. 🎉 👍

stories/features/shields/components/controls/scriptsControl.tsx
cezaraugusto: hey Ross thanks for the feedback. I'm not sure I'm following, we need the  `tabindex`  so the control row (includes the arrow down, stats, and text) could be tabbable so using the space bar would trigger the blocked resources list, as expected from an a11y perspective. 

the toggle is independent of that and has `focus` on click just because that's how the outline works, but we're talking about a default behavior for the `input[type="checkbox"]` element.

the first alternative I proposed was to have `focus` only in the arrow down icon, but Karen said that it's better to have in the row as it is now. if you think this is wrong, I can change it back or we can figure out another way, but I guess we'll need Karen's input as well

src/features/shields/structure.ts
cezaraugusto: sounds good but deferring to @karenkliu

rossmoody:
Duh, I was pressing enter and not spacebar. That's a great addition. No issue.

Just to be sure. This is what I'm seeing: https://d.pr/i/9kiate which seems tight in comparison to your Sketch doc: https://d.pr/i/SU6cWx  Fixed via: https://github.com/brave/brave-ui/pull/382#discussion_r258324587

tomlowenthal:
Will the production version support enter rather than space, or is it all space4lyf?

cezaraugusto:
@tomlowenthal disabled users should do their flow with the keys they expect to work which are <kbd>tab</kbd> for navigation, <kbd>space</kbd> for toggles and <kbd>space</kbd> and <kbd>Enter</kbd> for button actions. not usual to have <kbd>Enter</kbd> in other actions, although it can be implemented without any burden

tomlowenthal:
Thanks Cezar. I guess I'm not up on the keyboard-a11y norms. Definitely we should do whatever is the standard/expected thing. And thanks for showing me `<kbd>` markup. I didn't know about that before, and it's a neat technique.

karenkliu:
I dismissed my review but after going over these issues with Cezar, this PR addresses #2566, #1894, and #1892 - good to go. 👍

src/features/shields/media/arrowDown.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+import * as React from 'react'
+import GenerateComponentForGraphic from './template'
+
+// Direct access to SVG
+export const Graphic = (
+  <svg width='24' height='24' xmlns='http://www.w3.org/2000/svg' xmlnsXlink='http://www.w3.org/1999/xlink'>
+    <defs>
+      <path d='M12 13.262l4.762-4.081a.75.75 0 1 1 .976 1.138l-5.25 4.5a.75.75 0 0 1-.976 0l-5.25-4.5a.75.75 0 1 1 .976-1.138L12 13.262z'
+      id='a' />
```
petemill: Can we try to avoid ID attributes on inline elements, poluting global namespace. This will cause problems if there's multiple SVG files on the same page trying to do something with '#a' and '#b'. I beleive this is an issue from the sketch export, and it could either be fixed in sketch, or manually (which I've had to do before 😠  )

petemill: Also is this arrow different than our component arrow - if not, what's stopping you using the arrow component @cezaraugusto? cc @rossmoody

cezaraugusto:
after deliberating with @rossmoody and @petemill we're going to use the default brave-ui carat icon for now, which will be changed to something more closer to the original Shield spec defined TBD in a follow-up. this is what is looks like right now:

shot

since there are no other code comments I'm going to land this so we can quickly start porting everything to brave-core

PR #358, opened 2019-01-29
Color Rework & Refactor

rossmoody:
Fixes: https://github.com/brave/brave-ui/issues/363

This PR is meant to set the foundations for implementing theme variables more consistently across our components and get us on the road to Light/Dark theme adoption from the base component level. There is also a naming convention consistency refactor component to this which is making the amount of changes look bloated. I wanted to ensure a foundation and naming convention was squared away in one swoop.

Using some of the preliminary work and structure @petemill set up in a `button` component merge many months ago that incorporated a `Theme Provider`, this PR follows the same pattern with some additions to how color is generated and more uniform variable naming conventions. The `toggle` and `button` components were used as initial guinea pigs for implementing the new approach.

image

## High Level Changes
1. Moved `theme` and `color` assets inside `components/style`. This is to future proof the `style` umbrella for future code additions that might not necessarily fit under theme. This is illustrated by the `makeColor` function being broken out into the `function` folder under `style`. In my research I noticed that very often other libraries come to put motion and other style utilities in there.
2. Started `theme` variable foundations for quite a few components. The `button` variables were replaced though follow the same structure (just new names). A lot of the changes from this are the result of consolidating the them variable definitions across components down to the one theme file.
3. Added a systematic color palette generation mechanism. This is in the `style` folder. The naming convention for colors follows the same names and styles. The thing that updated was how they are generated. 
4. Generally I went through and tried to structure file names and folders in a consistent convention. This means making component folder names singular instead of plural and adopting common component names. Examples: `popupsModals` to `modal` | `buttonIndicators` to `button` | `dataTables` to `table` .

## Toggle
The toggle component received a style rework cause it wasn't importing theme variables and also was accommodating for `type?: 'dark' | 'light'` that will be swapped for `theme` color variables. 

## TinyColor
TinyColor is a TS dependency that performs basically the same color functions as Sass or Less. It's an integral part of generating states and consistent alphas in our theme files. It's also what is used to generate the color palette scales. https://github.com/TypeCtrl/tinycolor

## Theming Panel in Storybook
Though the only two components that can actually utilize this capability thus far are `Toggle` and `Button`, storybook has been set up to start implementing theming previews for all base components. This is through an addon that give a toggle to every story for going back and forth between themes. If the component isn't defined for a theme, it will default to `default`.

dark

## Background Colors in Stories
[...]

PR #384, opened 2019-02-21
WIP: BodyCopy, Footnote & Heading text components

rossmoody:
Here is a build to checkout the components in `Components/Text/` : https://brave-ui-meobs11es.now.sh/

This was built on the `color-rework` [branch](https://github.com/brave/brave-ui/pull/358) which is getting split up and merged by @petemill (ty btw 🥇 ) so this PR will need multiple rebases and reworks but it's a foundational start to establishing text components that leverage theme colors and variables.

Note: I started these as Class components to follow suit with brave ui but since they won't have a reason to leverage state (right?) I switched em over to func components. From what I'm reading on the interwebs this seems like a favorable approach?

There are examples and propTables in the stories section for each component but the general gist is:

- 3 components. BodyCopy, Footnote & Heading
- tag can be passed to any component to change from `p, h1, h2, div, span, etc`
- heading is the only component that can be passed `fontSize`. The sizes are changed in a sequential number scale from 0-9 that follow a modular type scale tied to theme. 
- BodyCopy is locked at 14px. This is to maintain consistency of body copy styling and size. 
- Footnote is locked at 12px. Same as above. 
- margin and padding are passed a css string via `0, 0, 0, 0`. they default to 0
- the color element of type will need tweaks and additions, but the way it meshes with theme is to adjust the alpha of the type to lighten it. in this way the exact same interaction will happen when it's reversed

| Prop | Type | Default | Value | Notes |
|--|--|--|--|--|
| fontSize | number | 4 | 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 | Scales via modular type scale. Only passed to heading. |
| tag | any |  | h1, h2, h3, h4, h5 |  |
| margin | string |  |  | Pass in CSS string: {'2px, 2px, 2px, 2px'} |
| padding | string |  |  | Pass in CSS string: {'2px, 2px, 2px, 2px'} |
| isItalic | boolean | false |  | Makes text italic |
| color | Color | default | default, defaultFade15, defaultFade30, defaultFade50 | Text color is adjusted via alpha fade to lighten. |
| children | React.ReactNode |  |  |  |
| lineHeight | LineHeight | default | default, large, zero | 1.5, 1.75, 0 |
| fontWeight | FontWeight | regular | regular, bold, light | 400, 600, 200 |


petemill:
This is awesome @rossmoody. 🍾 
I especially like having the `Body`, `Footnote`, `Heading` and `Text` as separate components since those names convey purpose and describe the content nicely.
My initial feedback would be to keep as much of that content-style separation as possible, namely:
- `color` prop - is there a semantic prop we could use for this, or at least some kind of simpler scale (e.g. 1, 2, 3, 4, 5) instead of explicitly having to pass through `defaultScale{15, 30, 50}`?
- `margin` and `padding` as props on these components: this requires putting margin and padding values inline in a react component, instead of in a layout's style definition. Can we put these at `0` and allow an element to either use a wrapper, or to extend via styled-component's extend function.
- `isItalic` prop would be better as `emphasis` IMO so you could have `<BodyCopy emphasis>Wassup</BodyCopy>`. Right now we use italic to style emphasised text...
- `fontWeight` - perhaps we could use something like a `strongLevel` with values like 0, 1, 2.
- `lineHeight` - what are the UX scenarios that the designer would choose different line heights? In the _Heading_ component, could lineHeight be tied to the size prop so it would control both font-size and line-height, or is that not how we are designing?

rossmoody:
Changes would be too plentiful to keep this open and since it's not against master gonna close it but feedback was great and any other feedback is welcome.

PR #387, opened 2019-02-22
1: Storybook and package.json update

rossmoody:
The first installation in breaking up the great https://github.com/brave/brave-ui/pull/358 monster. 

- update package.json to include TinyColor dependency
- update storybook addon order from most used to least in side panel
- moved addon panel to the right
- webpack trailing semi colons deleted
- added github styling css file for future implementation of Markdown styling in stories instead of depending on unofficial storybook addons
- ommitted manager-head.html styling as addons now override them
- removed old components and stories from view in storybook


.storybook/style/github_style.css
```diff
@@ -0,0 +1,961 @@
...
+.md h2:hover .anchor .octicon-link,
+.md h3:hover .anchor .octicon-link,
+.md h4:hover .anchor .octicon-link,
+.md h5:hover .anchor .octicon-link,
+.md h6:hover .anchor .octicon-link {
+  visibility: visible;
+}
+
+.md {
+  -
```
cezaraugusto: this accidental `-` is breaking the build

rossmoody: fixed up.

PR #388, opened 2019-02-23
2: Folder refactor

rossmoody:
The second installation of breaking up the https://github.com/brave/brave-ui/pull/358 monster. No code changes in the component section of this. Only a refactor of the folder structure and tidying the stories section to match.


rossmoody:
still too big i think. gonna break it apart again.

PR #389, opened 2019-02-25
3: Color palette and theme introduction

rossmoody:
## Color palette and theme foundation introduction
- introduced a general tentative framework for theme variables
- introduced the makeColor function for generating palette
- updated color palette values and names

##  Button updates
- adjusted sizing to be more in line with latest design spec
- adjusted default button color to be lighter gray to replace subtle (eventually)
- adjusted 'warn' to 'danger' prop and synced to theme
- gave the states a transition

## Toggle refresh
- adjusted styling and general css and linked to theme
- removed inline theme reference of 'light'/'dark'
- aligned size naming convention to 'default ' | 'small'
- additional state polish

## Theme linking
- linked existing component reference to correct theme variables
- sync fix where button sizing was stacking cause of size change


rossmoody:
still too many separate categories of change in this. gonna break it apart.

PR #390, opened 2019-02-27
Update to V4 styled-components

rossmoody:
Closes: https://github.com/brave/brave-ui/issues/128

Main driver for this was to update `styled-components` to v4 so we can leverage the `as` polymorphic prop so I could use it in my typography component experimentation. Of course you can't just update one thing cause then everything has to come with it so I'll outline updates and their drivers:

### styled structure

Across the board the new `styled-component` structure was updated:
```
V4
styled.div<Props>

V3
styled<Props, 'div'>('div)
```

### innerRef -> ref

With the updates I was getting errors on instances of `innerRef` instances and after [reading it was deprecated](https://www.styled-components.com/docs/api#deprecated-innerref-prop) the instances were switch to just `ref`

## Module Paths
All instances of `styled-component` module imports were run through the `components/src/style/theme` path instead of direct to `styled-components`.

## Dependency Updates

1. `styled-components": ^4.1.3` - latest styled components release
2. ` @types/styled-components: ^4.1.10`: adds necessary TS definitions for v4
3. `typescript-plugin-styled-components: 1.2.0` - wouldn't work without it
4. `jest-styled-components": "^6.3.1" | "enzyme-adapter-react-16": "^1.10.0", |  "react-test-renderer": '^16.8.3"` - snapshots were breaking and not inheriting styles with v4. all 3 of these needed to be updated to get it to inherit the styled-component styles in the snapshot.
5. ts.config ->  `"types": ["react", "jest"], // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33311`: the "@types/styled-components" dependency (for some reason) inherits a react-native type definition folder. styled-components hates it and it triggers an error for duplicate type definitions. this fixes that error. read more: https://github.com/DefinitelyTyped/DefinitelyTyped/issues/33015

This should probably come after the new shields PR merge as there was some wonky styled-component definitions errors going on in `stories/features/shields/index.tsx` and since it's getting an overhaul I just made it work by omission  instead of troubleshooting through it. (i.e. `stories/features/shields/index.tsx`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -52,7 +52,6 @@
       "version": "7.0.0",
       "resolved": "https://registry.npmjs.org/@babel/helper-annotate-as-pure/-/helper-annotate-as-pure-7.0.0.tgz",
       "integrity": "sha512-3UYcJUj9kvSLbLbUIfQTqzcy5VX7GRZ/CCDrnOaZorFFM01aXp1+GJwuFGV4NDDoAS+mOUyHcO6UD/RfqOks3Q==",
-      "dev": true,
       "requires": {
         "@babel/types": "^7.0.0"
       }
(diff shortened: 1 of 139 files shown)


rossmoody:
PR'd this against refactor adjustments i made. easier to start fresh.

PR #407, opened 2019-03-03
Update theme story styling

rossmoody:
Just some general storybook polish. No issue tied to this one. Added a hex value to the cards. Removed the chapters addon reference as we are going to move to documentation outside of storybook in some capacity.

## Before
screen shot 2019-03-02 at 4 05 47 pm

## After

screen shot 2019-03-02 at 4 23 34 pm

### DIFF
--- a/stories/components/theme.tsx
+++ b/stories/components/theme.tsx
@@ -3,49 +3,58 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import * as React from 'react'
-// Storybook requires
-import { storiesOf, setAddon } from '@storybook/react'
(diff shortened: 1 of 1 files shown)


stories/components/theme.tsx
```diff
@@ -1,49 +1,58 @@
...
 
-// wrappers
+const Palette = styled.div`
+  display: grid;
+  grid-gap: 8px;
+  grid-template-columns: repeat(10, 1fr);
+  @media only screen and (max-width: 1300px) {
+    grid-template-columns: repeat(5, 1fr);
+  }
+` as any
```
cezaraugusto: this is ok for now as other components here use it but let's keep track of `any` next time we revist this file and remove all

PR #419, opened 2019-03-08
Shields

cezaraugusto:
This is a revert of shields revert.

Do not merge until brave-core changes land.


cezaraugusto:
merging as this was approved before but reverted in master to avoid conflicts

PR #426, opened 2019-03-13
Shields for brave-core 0.63.x

cezaraugusto:


src/features/shields/display.ts
```diff
@@ -0,0 +1,189 @@
...
+  line-height: 1.2;
+`
+
+interface ToggleStateTextProps {
+  status: 'enabled' | 'disabled'
+}
+
+export const ToggleStateText = styled<ToggleStateTextProps, 'span'>('span')`
+  box-sizing: border-box;
+  color: ${p => p.status === 'enabled' ? palette.orange400 : palette.grey300};
```
rossmoody: The `Down` color is a little light here and isn't passing AA color contrast checker so we should update since it's an important word. Anything grey700+ passes: https://d.pr/i/n1Onha

`color: ${p => p.status === 'enabled' ? palette.orange400 : palette.grey700};`

src/features/shields/select/carat.svg
```diff
@@ -1,14 +1,4 @@
...
-      <use xlink:href="#a"/>
-    </mask>
-    <use fill="#4A4A4A" fill-rule="nonzero" xlink:href="#a"/>
-    <g mask="url(#b)" fill="#E9E9F4">
-      <path d="M0 0h24v24H0z"/>
-    </g>
-  </g>
+<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 32 32' style='transform: rotate(270deg);width:20px;height:20px;'>
+    <path d='M22.4 29a1 1 0 0 1-.67-.25l-13.4-12a1 1 0 0 1 0-1.5l13.4-12a1 1 0 0 1 1.33 1.5L10.5 16l12.56 11.25A1 1 0 0 1 22.4 29z'
+    />
```
rossmoody: A little inconsequential but if you wanted to ditch the viewbox and transforms here could swap in this: 
```
<svg width="32" height="32" xmlns="http://www.w3.org/2000/svg"><path d="M16 19.047l11.04-9.2a1.5 1.5 0 0 1 1.92 2.305l-12 10a1.5 1.5 0 0 1-1.92 0l-12-10a1.5 1.5 0 1 1 1.92-2.304l11.04 9.2z"/></svg>
```

ref: https://d.pr/i/ZVfrCo

src/features/shields/display.ts
rossmoody: My B. This ship has sailed, just realized the perogative for this is different than what I'm calling out. Will file issue for only this.

cezaraugusto:
btw this was reverted here https://github.com/brave/brave-ui/commit/0f2221e4b5f153bb1070c9ffe22f3c88b52c0d68 as we won't make shields in 0.63.x

PR #432, opened 2019-03-19
Shields improvements

cezaraugusto:
* do not add opacity to shields row text when disabled
* do not wrap urls with queries in static list
* allow selection of blocked resources in the static list
* trim long site names

`now` link with changes: https://brave-ui-9qm8j8br4.now.sh

### DIFF
--- a/src/features/shields/structure.ts
+++ b/src/features/shields/structure.ts
@@ -278,6 +278,7 @@ export const BlockedListItem = styled<{}, 'li'>('li')`
   padding: 9px 0px;
   line-height: 1;
   white-space: nowrap;
+  user-select: all;
 `
 
 export const BlockedListItemWithOptions = styled<{}, 'li'>('li')`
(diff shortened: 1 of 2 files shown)


src/features/shields/structure.ts
```diff
@@ -141,7 +141,10 @@ export const BlockedInfoRowData = styled<BlockedInfoRowDataProps, 'div'>('div')`
   grid-gap: 2px;
   align-items: center;
   pointer-events: ${p => p.disabled && 'none'};
-  opacity: ${p => p.disabled && '0.4'};
+
+  & > *:not(:last-child) {
```
petemill: Will the text always be the last-child, or can we refer to the actual element selector, e.g.: 

```
& > * { opacity: ... }
& > ${StyledText} { opacity: 1 }
```

nit: Do you need the & if it's at the beginning of the line? At least it's explicit

cezaraugusto: without the `:not` operator stylelint starts complaining on `no-descending-specificity` for a totally unrelated style in `MainToggle`. 

```
 149:3  ✖  Expected selector ".selector73 > *" to come before selector ".selector66 > *:last-child"   no-descending-specificity
```

PR #434, opened 2019-03-21
Using HTML select for walletPanel

ryanml:
## Changes

Using plain `<select>` for WalletPanel to avoid clipping issues. Should maintain the look and feel of the previous component.

Screen Shot 2019-03-20 at 5 14 53 PM

## Test plan

https://brave-ui-gz4ablrx5.now.sh

## Integration
  
### DIFF
--- a/src/features/rewards/walletPanel/__snapshots__/spec.tsx.snap
+++ b/src/features/rewards/walletPanel/__snapshots__/spec.tsx.snap
@@ -11,7 +11,7 @@ exports[`WalletPanel tests basic tests matches the snapshot 1`] = `
 
 .c1 {
   min-height: 250px;
-  padding: 25px 30px 25px 30px;
+  padding: 15px 30px 20px 30px;
 }
 
 .c9 {
--- a/src/features/rewards/walletPanel/index.tsx
+++ b/src/features/rewards/walletPanel/index.tsx
@@ -24,13 +24,14 @@ import {
   StyleToggleTips,
   StyledNoticeWrapper,
   StyledNoticeLink,
-  StyledProfileWrapper
+  StyledProfileWrapper,
+  StyledSelect
 } from './style'
 
 // Components
-import { Select, Toggle } from '../../../components'
+import { Toggle } from '../../../components'
 
-import { RewardsButton, Tokens } from '../'
+import { RewardsButton } from '../'
 import ToggleTips from '../toggleTips/index'
 import Profile, { Provider } from '../profile/index'
 
@@ -53,7 +54,7 @@ export interface Props {
   toggleTips?: boolean
   donationAction: () => void
   onToggleTips: () => void
-  onAmountChange: () => void
+  onAmountChange: (event: React.ChangeEvent<HTMLSelectElement>) => void
   onIncludeInAuto: () => void
   onRefreshPublisher?: () => void
   refreshingPublisher?: boolean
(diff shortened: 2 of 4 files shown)


src/features/rewards/walletPanel/__snapshots__/spec.tsx.snap
```diff
@@ -11,7 +11,7 @@ exports[`WalletPanel tests basic tests matches the snapshot 1`] = `
 
 .c1 {
   min-height: 250px;
-  padding: 25px 30px 25px 30px;
+  padding: 15px 30px 20px 30px;
```
ryanml: Tightens up the wallet panel padding

src/features/rewards/walletPanel/style.ts
```diff
@@ -107,3 +107,21 @@ export const StyledNoticeLink = styled<StyleProps, 'a'>('a')`
...
 `
+
+export const StyledSelect = styled<StyleProps, 'select'>('select')`
+  width: 100%;
+  background: inherit;
+  height: 34px;
+  font-size: 14px;
+  border: none;
+  text-align: right;
+  color: #696fdc;
```
NejcZdovc: can we use theme color?

ryanml: We can, I just pulled this value directly from the `Tokens` component

src/features/rewards/walletPanel/index.tsx
```diff
@@ -90,25 +97,21 @@ export default class WalletPanel extends React.PureComponent<Props, {}> {
 
     return (
       <StyledSelectWrapper>
-        <Select
-          floating={true}
-          showAllContents={true}
+        <StyledSelect
           value={monthlyAmount}
-          onChange={this.props.onAmountChange}
+          onChange={this.onChange}
```
NejcZdovc: do we need this change? Can't we just handle it as prop passed in and then in core we do `event.target.value`? Don't see real value in local `onChange` function that just calls props function

src/features/rewards/walletPanel/style.ts
NejcZdovc: let's use theme color, so that it will be easier to do theming

src/features/rewards/walletPanel/__snapshots__/spec.tsx.snap
NejcZdovc: 👍

src/features/rewards/walletPanel/index.tsx
ryanml: removed

src/features/rewards/walletPanel/index.tsx
```diff
@@ -90,25 +91,27 @@ export default class WalletPanel extends React.PureComponent<Props, {}> {
...
-          floating={true}
-          showAllContents={true}
+        <StyledSelect
           value={monthlyAmount}
           onChange={this.props.onAmountChange}
         >
           {donationAmounts.map((token: Token, index: number) => {
+            const tokenValue = token.tokens.toString()
+            const paddingLength = tokenValue.length < 5
+              ? tokenValue.length === 4 ? 3 : 4
```
ryanml: The number of space entities that look best for values of length 4 and 3 are 3 and 4 respectively. It didn't end up just being some math on the length - 5.

PR #438, opened 2019-03-28
Unverified publishers in panel can now have status refreshed

jasonrsadler:
Addresses https://github.com/brave/brave-browser/issues/3339

## Changes
Added 'Check again...' link to check for publisher's verified status

## Test plan
Open unverified wallet panel, verify 'Check again...' link
image

- https://storybook-out.jasonrsadler.now.sh?path=/story/feature-components-rewards-wallet-desktop--panel

## Integration
  
### DIFF
--- a/src/features/rewards/profile/index.tsx
+++ b/src/features/rewards/profile/index.tsx
@@ -15,10 +15,12 @@ import {
   StyledProviderWrap,
   StyledInlineVerified,
   StyledVerifiedText,
-  StyledInlineUnVerified
+  StyledInlineUnVerified,
+  StyledVerifiedCheckLink,
+  StyledRefresh
 } from './style'
 import { getLocale } from '../../../helpers'
-import { VerifiedIcon, UnVerifiedIcon } from '../../../components/icons'
+import { VerifiedIcon, UnVerifiedIcon, LoaderIcon } from '../../../components/icons'
 
 export type Provider = 'twitter' | 'youtube' | 'twitch'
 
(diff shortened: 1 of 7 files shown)


ryanml:
@jasonrsadler can you generate a `now.sh` link for this feature? Like https://brave-ui-m1yk2rzha.now.sh/

src/features/rewards/walletPanel/index.tsx
```diff
@@ -55,6 +55,7 @@ export interface Props {
   onToggleTips: () => void
   onAmountChange: () => void
   onIncludeInAuto: () => void
+  onRefreshPublisher: () => void
```
NejcZdovc: this one needs to be optional

jasonrsadler: Done.

src/features/rewards/profile/index.tsx
```diff
@@ -109,6 +117,18 @@ export default class Profile extends React.PureComponent<Props, {}> {
               <StyledVerifiedText>
                 {getLocale('unVerifiedPublisher')}
               </StyledVerifiedText>
+              {
+                !publisherRefreshed ?
```
NejcZdovc: we don't need new prop, we can use `onRefreshPublisher` to determinate if we should show it or not

jasonrsadler: The problem I'm having with that is that we also need value to determine whether spinner is shown. Open to ideas.

NejcZdovc: talked it over on the call, so all good here

PR #463, opened 2019-05-06
Fix dropdown colors for Shields in Windows

cezaraugusto:
Test Plan:

Ensure dropdowns on Shields have visible text across platforms

Now link: https://brave-ui-e45cj4qm1.now.sh

### DIFF
--- a/src/features/rewards/walletSummary/__snapshots__/spec.tsx.snap
+++ b/src/features/rewards/walletSummary/__snapshots__/spec.tsx.snap
@@ -185,7 +185,7 @@ exports[`WalletSummary tests basic tests matches the snapshot 1`] = `
     <div
       className="c3"
     >
-      MISSING: monthApr
+      MISSING: monthMay
        
       2019
     </div>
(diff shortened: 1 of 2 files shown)


src/features/rewards/walletSummary/__snapshots__/spec.tsx.snap
```diff
@@ -185,7 +185,7 @@ exports[`WalletSummary tests basic tests matches the snapshot 1`] = `
     <div
       className="c3"
     >
-      MISSING: monthApr
+      MISSING: monthMay
```
petemill: 👍 🤦‍♀

src/features/shields/select/index.ts
```diff
@@ -41,7 +41,8 @@ export const SelectBox = styled<SelectBoxProps, 'select'>('select')`
   cursor: pointer;
 
   > option {
-    color: ${palette.white};
+    /* DO NOT MODIFY: this is for the dropdown only and has effect only on Windows */
```
petemill: I'm thinking 'Do not modify' is not helpful enough. At some point this will be modified for one reason or another, and it will be useful to see _why_, within the way the code is currently, this needs to be set as 'black' for windows. Perhaps just pointing to the github issue link will be enough to set the context @cezaraugusto ?

cezaraugusto: sounds good, updated

PR #468, opened 2019-05-16
sanitize social media links in tipping banner

### LINKED ISSUE #466: Tipping banner should sanitize social media links
Hackerone issue: https://hackerone.com/reports/577426
Discussion thread in Slack: https://bravesoftware.slack.com/archives/G2KN13Z8C/p1557962234053000
Publishers issue: https://github.com/brave-intl/publishers/issues/1891

diracdeltas:
fix #466

## Test plan
1. go to https://www.nate.blue/ (it may not load; that's fine)
2. click rewards icon to send a tip
3. no social icons should appear

- No visual changes

## Integration
  
### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,8 @@
       "!src/**/index.ts"
     ],
     "testMatch": [
-      "<rootDir>/src/**/spec.tsx"
+      "<rootDir>/src/**/spec.tsx",
+      "<rootDir>/src/*.test.ts"
     ],
     "testPathIgnorePatterns": [
       "<rootDir>/node_modules/"
--- a/src/features/rewards/siteBanner/index.tsx
+++ b/src/features/rewards/siteBanner/index.tsx
@@ -36,7 +36,7 @@ import {
 
 import Donate from '../donate/index'
 import Checkbox from '../../../components/formControls/checkbox/index'
-import { getLocale } from '../../../helpers'
+import { getLocale, normalizeSocialUrl } from '../../../helpers'
 import {
   CloseCircleOIcon,
   TwitterColorIcon,
@@ -121,6 +121,10 @@ export default class SiteBanner extends React.PureComponent<Props, State> {
 
     const self = this
     return social.map((item: Social) => {
+      const href = normalizeSocialUrl(item.type, item.url)
+      if (!href) {
+        return null
+      }
       const logo = self.getSocialData(item)
       return (
         <StyledSocialItem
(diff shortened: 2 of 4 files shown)


diracdeltas:
@dlipeles i'd rather just normalize item.url since it's used directly as an href attribute - so for instance if `item.url` is `www.twitch.tv/test`, i would normalize that to `https://twitch.tv/test` before it's set as an href.

diracdeltas:
> @dlipeles i'd rather just normalize item.url since it's used directly as an href attribute - so for instance if `item.url` is `www.twitch.tv/test`, i would normalize that to `https://twitch.tv/test` before it's set as an href.

this can be done here in this PR

diracdeltas:
updated the implementation and added unit tests which passed locally. i tried to do the manual test plan but nate.blue doesn't show up as verified in Brave dev

NejcZdovc:
@diracdeltas Travis is failing, can you please check why https://travis-ci.org/brave/brave-ui/jobs/533611667

petemill:
I don’t want to delay this since this implementation will work, but do you think a better place for this kind of logic is in the (c++) layer that fetched the data and provides it to the ui layer, rather than inside a UI display component? @nejczdovc @diracdeltas

diracdeltas:
> will work, but do you think a better place for this kind of logic is in the (c++) layer that fetched the data and provides it to the ui layer

I'm a bit mixed on this. I think logically it makes sense in the C++ layer, but it is easier to ensure these URLs are always sanitized if the sanitization checks happens right before the href is set. I guess it's a question of whether it's acceptable for the UI layer to blindly trust that the input from the C++ layer is sanitized.

 @NejcZdovc thanks, i'll work on fixing those

petemill:
> I'm a bit mixed on this. I think logically it makes sense in the C++ layer, but it is easier to ensure these URLs are always sanitized if the sanitization checks happens right before the href is set. I guess it's a question of whether it's acceptable for the UI layer to blindly trust that the input from the C++ layer is sanitized.

Thanks - in that case @diracdeltas I would suggest if keeping it in the UI layer (the typescript), that it would be better to put it in the data-retrieval part, just after the URLs come back from c++ and are added to the state. As you have it now, it is in the render layer, which will re-run often and unnecessarily, and we want to avoid this performance-problematic pattern.

diracdeltas:
@petemill can you point me at where that would be? i assume it's in brave-core?

diracdeltas:
closing in favor of b-c PR

Issue #467, opened 2019-05-16
npm audit failure


diracdeltas:
```
found 2 moderate severity vulnerabilities
  run `npm audit fix` to fix them, or `npm audit` for details
```

running npm audit fix shows: `2 vulnerabilities required manual review and could not be updated`

Issue #454, opened 2019-04-19
Remove storybook chapters addon


rossmoody:
Last two or three times it caused a security problem since updating to v4 Storybook. It's only leveraged on buttons page and since that documentation will live in design.brave we can omit it to save some hassle.

Issue #451, opened 2019-04-19
Remove large static images from Rewards stories


rossmoody:
Should probably omit the large background images from the rewards wallet stories. Has a tendency to make using the developer panel with absolutely positioned elements so far right much more difficult and it's a rather old browser mockup.

Issue #450, opened 2019-04-19
Remove Bart Baker from Storybook examples


rossmoody:
Be gone Bart. Need to opt for something a little more Brave/BAT-ish as our example in storybook since the screenshots are getting posted repeatedly in public forums.

PR #474, opened 2019-05-22
Storybook housekeeping 

rossmoody:
Resolves https://github.com/brave/brave-ui/issues/450
Resolves https://github.com/brave/brave-ui/issues/451
Resolves https://github.com/brave/brave-ui/issues/454
Resolves https://github.com/brave/brave-ui/issues/467


stories/assets/locale.ts
```diff
@@ -208,9 +208,9 @@ const locale: Record<string, string> = {
   turnOnRewardsTitle: 'Turn on Rewards',
   type: 'Type',
   uhOh: 'Uh oh…',
-  unVerifiedCheck: 'Check Again...',
-  unVerifiedPublisher: 'Not yet verified',
-  unVerifiedText: 'This creator has not yet signed up to receive contributions from Brave users.',
+  unVerifiedCheck: 'Refresh status',
```
jasonrsadler: Keep in mind when you change the text values, it only changes for storybook and won't reflect the current string usage in `brave-core`

stories/assets/locale.ts
```diff
@@ -9,7 +9,7 @@ const locale: Record<string, string> = {
   activityNote: 'To protect your privacy, this Brave Rewards statement is not saved, recorded or logged anywhere other than on your device (this computer). It cannot be retrieved from Brave in the event of data loss on your device.',
   addFunds: 'add funds',
   addFundsFAQ: 'the FAQ',
-  addFundsNote: 'Reminder: The Brave Wallet is unidirectional and BAT flows to publisher sites. For more information about Brave Payments, please visit',
+  addFundsNote: 'Reminder: The Brave Wallet is unidirectional and BAT flows to Creators\' sites. For more information about Brave Rewards, please visit',
```
jasonrsadler: "creator" is a common noun. Unless there is a design specification that says otherwise, we should lowercase it.

stories/assets/locale.ts
rossmoody: yeah good spot. These should prob not be mixed in here but i cherry picked. i'll be sure to make them a part of the core sha pr

stories/assets/locale.ts
rossmoody: this should definitely be tackled shortly on the design language initiative. could punt for now but the thinking is in the context of Brave products we want to brand them as Brave Creators instead of just content creators. i.e. they aren't just another creator they are a Creator.

jasonrsadler: Of course. Desing language not being my arena, but in my personal opinion, I would consider using `Brave Creators` specifically and `creators` when the brand is not included

rossmoody: that's safer for the interim, i'll amend

PR #478, opened 2019-05-25
Grammatical Correction

romant:
Updating the usage of the verb

## Changes
- `e` swapped in for `a`

## Test plan
- Only if string matching breaks downstream screen validation. Could not find any other references, as such should have no impact on test validity

- See [affect vs effect](https://en.oxforddictionaries.com/usage/affect-or-effect) in the Oxford Dictionary


rossmoody:
Thank you and good spot!

Reference: https://github.com/brave/brave-browser/issues/4088

PR #484, opened 2019-05-29
Don't vertically clip creator header images in tip UI

### LINKED ISSUE #483: Remove mobile styling configuration from SiteBanner
We used to use the `siteBanner` component within brave-ui for Rewards on Android, but this is now a native component. We can go ahead and remove the mobile stylings made for it, `isMobile` and references to it can be safely removed.

cg505:
## Changes
See brave/brave-browser#2015. The banner image for the tip interface now scales height dynamically.
image
image

Also resolves #483 and #488.

### With header image
- Launch with `--rewards=staging=false`. 
- Visit brave.com and make sure browser width is at least 1500px or so.
- Open rewards box. If it says "Not yet verified" click "Check again...".
- Click "SEND A TIP..." button.
- The header image should say "Brave the new internet. Fast, safe, secure, corageous." Make sure this image is not vertically clipped.
- Try again at a small browser width (~1000px). It should work as before.
### Without header image
- Visit any site *without* a verified publisher and make sure browser width is at least 1500px or so.
- Open rewards box and click "SEND A TIP..." button.
- The header should look normal and should be 176px tall.

https://brave-ui-6cuwjq0t3.now.sh?path=/story/feature-components-rewards-concepts-desktop--site-banner (toggle "Show bg image" knob)

## Integration
  

stories/features/rewards/concepts.tsx
```diff
@@ -466,61 +466,3 @@ storiesOf('Feature Components/Rewards/Concepts/Mobile', module)
     />
   ))
   .add('Settings', () => <SettingsMobile />)
-  .add('Site Banner', withState({ donationAmounts, currentAmount: '5.0', showBanner: true }, (store) => {
```
NejcZdovc: why was this one removed?

cg505: It is only for mobile, which was removed. (notice the `isMobile={true}` prop)

ryanml: Can confirm, this is fine to remove 👍

src/features/rewards/siteBanner/index.tsx
```diff
@@ -287,20 +284,19 @@ export default class SiteBanner extends React.PureComponent<Props, State> {
                     </StyledNoticeWrapper>
                     : null
                 }
-                <StyledTitle isMobile={isMobile} isTwitterTip={(screenName && screenName !== '') ? true : false}>
+                <StyledTitle isTwitterTip={(screenName && screenName !== '') ? true : false}>
```
ryanml: I know this ternary was inline before, but can we cache it in to a variable this rendering? Would look a lot cleaner :)

cg505:
@ryanml Actually all the banner images are the same aspect ratio, so the existing knobs are sufficient. The only thing to check is the different widths.

dlipeles:
> all the banner images are the same aspect ratio, so the existing knobs are sufficient. The only thing to check is the different widths.

I can attest to this, we store all of our images server side at 2700 x 528 👍 

That said, things could change in the future, might not hurt to create more knobs just for testing purposes.

NejcZdovc:
One thing that we need to keep is height of the image to be fixed, otherwise height will be too big and small screens will have problems. So what we need to achieve is, with height that we have defined, how to best fit it in there.

cg505:
@NejcZdovc Since the aspect ratio is so wide (900 by 176) you will have to have a really wide and really short screen for it to be a problem (like 2 or 3x as wide as tall). If we're worried about it, we could either cap the width of the image or vertically center it past a certain point. Not sure what the best approach would be in that caes.

cg505:
The dialog box height is [capped at 700px](https://github.com/brave/brave-core/blob/3095dfc1078ac4e854388b5345aeaf88b24d93e1/browser/brave_rewards/tip_dialog.cc#L35). We can add a max-height to the banner image so that everything should fit even if your screen is super wide.

cg505:
Pushed a change to cap the overally height of the banner image at 350px. This will avoid creating a scrollbar even on really wide screens. Looks like this:
image

Do we want to handle this case some other way as well (like max-width or vertical centering)?

dlipeles:
Is there anyway we can handle these images in a way that doesn't cut out, especially important if the image contains text @NejcZdovc @cg505

cg505:
@dlipeles When the screen is really wide, we could somehow restrict the width of the banner image, but I'm not sure exactly how that will work. (What should be the background color? Should we restrict the width of the entire dialog?)

dlipeles:
I think setting a max-width on the entire banner might be a good idea, I will defer to @NejcZdovc

src/features/rewards/siteBanner/style.ts
```diff
@@ -60,13 +60,20 @@ export const StyledBannerImage = styled<Partial<Props>, 'div'>('div')`
   font-weight: 600;
   line-height: 0.74;
   color: #d1d1db;
-  height: 176px;
+  min-height: 176px;
+  max-height: 350px;
```
petemill: @cg505 It's not clear where this number 350px comes from? Might be uncomfortable to change in the future if we don't know why it's set at that number.

cg505:
Enforced a maximum width in the brave/brave-core#2566 so that the maximum height is no longer needed.

src/features/rewards/walletSummary/spec.tsx
```diff
@@ -25,9 +25,16 @@ describe('WalletSummary tests', () => {
 
   describe('basic tests', () => {
     it('matches the snapshot', () => {
+      const RealDate = Date
+      global.Date = class extends RealDate {
+        constructor () {
```
petemill: 👍

src/features/rewards/siteBanner/style.ts
cg505: this got removed with the change to brave-core

PR #486, opened 2019-05-31
Button focus styling

rossmoody:
Addresses https://github.com/brave/brave-browser/issues/4658

This one is just a revision of the black focus squares we have going on in the default buttons at the moment. This can be reverted when the shared button component is more thoroughly defined.

These changes will have to happen in large part (I believe) in Core repo. The styling approach took into consideration the theming update in the neighboring PR.

### DIFF
--- a/src/features/welcome/button/index.ts
+++ b/src/features/welcome/button/index.ts
@@ -3,7 +3,8 @@
  * You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 import { ComponentType } from 'react'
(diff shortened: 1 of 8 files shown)


rossmoody:
This one will need some updates on the core side as well. I started a branch but was having build issues. @cezaraugusto want me to do the core or or you got it?

cezaraugusto:
could you try again using current latest? if error persist feel free to DM and we can go over together

PR #497, opened 2019-06-11
Uphold wallet verification UI

cg505:
## Changes

Adds UI components (button at top, footer updates, verification panel screen) to indicate uphold wallet status.

See brave/brave-browser#4774.

- https://brave-ui.ccooper1.now.sh/?path=/story/feature-components-rewards-concepts-desktop--wallet-panel (disable "show notification" in knobs, enable "show uphold" and then play with "wallet status")
- https://brave-ui.ccooper1.now.sh/?path=/story/feature-components-rewards-other-desktop--panel-verification

## Integration
- [x] Does this contain changes to src/components or src/
  - [x] Will you publish to npm immediately after this PR, or wait until sometime in the future?
  - [x] Adding new backwards-compatible functionality? _(minor version increase)_
  
  - [x] Will you create brave-core PR to update to this commit after it is merged?


src/components/buttonsIndicators/button/index.tsx
```diff
@@ -22,6 +22,8 @@ export interface Props {
   id?: string
   disabled?: boolean
   icon?: {image: React.ReactNode, position: 'before' | 'after'}
+  beforeIcon?: React.ReactNode,
```
NejcZdovc: does the existing code not work for your case? in `icon` object you already have `before` and `after` options

cg505: I need both `before` and `after` at the same time, which is why I changed this.

NejcZdovc: could we use existing icon object to achieve this and just extend it? I would like to avoid duplicating props.

cg505: I couldn't think of a good way to extend the current `icon` prop to work for two different icons, so I thought this would be the cleanest solution. But if you have a good idea for how to do it inside `icon` that totally works.

src/features/rewards/walletPopup/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+    } = this.props
+    return (
+      <StyledWrapper onClick={onClose} id={id}>
+        <StyledDialog onClick={this.insideClick}>
+          <StyledContent>
+            <StyledHeader>
+              <StyledIcon>
+                <UpholdColorIcon />
+              </StyledIcon>
+              {'@' + username}
```
ryanml: Let's maybe do a quick check for `username.length` so we don't potentially display just an `@`

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -318,6 +335,98 @@ export default class WalletWrapper extends React.PureComponent<Props, State> {
...
+                <StyledLink onClick={onVerifyClick}>
+                  {getLocale('walletReconnect')}
+                </StyledLink>
+              </li>
+            </StyledDialogList>
+            : <>
+              <StyledLink onClick={onVerifyClick}>
+                {getLocale('walletVerifyLink')}
+              </StyledLink>
+              {' '}
```
ryanml: Is this for stylistic purposes? Can we maybe accomplish this via CSS if so?

src/features/rewards/panelVerify/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+  )
+
+  render () {
+    const {
+      onVerifyClick,
+      onClose,
+      id
+    } = this.props
+
+    return (
```
ryanml: Could we break this up in to a couple more methods? Maybe one for `footer` and `header`

src/features/rewards/walletWrapper/index.tsx
cg505: It's not really for stylistic purposes, there should just be a space between the link and the rest of the test and the space will get eaten up by JSX otherwise.

src/features/rewards/walletPopup/index.tsx
cg505: If `username.length === 0` what should we do? We probably shouldn't be showing the box at all in this case

cg505: Actually we do this check in `WalletWrapper`, which will only display the box if `username` is truthy. and `''` is not truthy.

src/features/rewards/panelVerify/index.tsx
cg505: done

src/components/buttonsIndicators/button/index.tsx
NejcZdovc: cc @petemill for his thoughts

NejcZdovc:
let's not merge until core PR is ready

src/components/buttonsIndicators/button/index.tsx
petemill: This might need some design decision to clarify the change to the button in the system - @rossmoody? I'm not sure it's easily understandable for the user to grasp the meaning of the button when there are 2 icons.
image

If it isn't a button, but is some kind of 'status chip' then let's make a new component called that. It could for example have a prop which determines which icon is displayed called `status` which accepts `'success' / 'fail' / 'info'`

rossmoody: This is the first I'm seeing this and sadly I really don't have any insight as to how this works or what function it serves. @jenn-rhim will need to chime in on details surrounding this.

jenn-rhim: This is not the approved design. We removed this state. @NejcZdovc

cg505:
Removed "Connected" state and rebased everything, ready for re-review.

src/features/rewards/walletPopup/index.tsx
```diff
@@ -0,0 +1,62 @@
...
```
NejcZdovc: let's remove @ as full name will be provided by c++

src/features/rewards/walletWrapper/index.tsx
```diff
@@ -130,19 +140,24 @@ export interface Props {
   onFinish?: () => void
   onSolution?: (x: number, y: number) => void
   convertProbiToFixed?: (probi: string, place: number) => string
+  onVerifyClick?: () => void
+  goToUphold?: () => void
+  username?: string
```
NejcZdovc: nit: change to `userName`

src/components/buttonsIndicators/button/index.tsx
NejcZdovc: @petemill @rossmoody @jenn-rhim so even though that that specific button was not needed we still need it for this one:

image

PR #504, opened 2019-06-24
✨New welcome flow for importing settings from other browsers

imptrx:
Issue tracker: https://github.com/brave/brave-browser/issues/1530
Spec: https://github.com/brave/brave-core/pull/2679#issuecomment-501847113

## Changes
- Inline import option in the welcome flow for importing settings from other browsers
- **Dev Concern**: I copied the `Select` into "features/welcome" since we have multiple "screens" in welcome relying on it - it didn't make sense to have all of them cross-referencing a shields component (I am assuming we can lift it to a general directory when we migrate our stories to core)
- Quick view:
welcome-import

## Test plan
- `brave://welcome` should still load properly when loaded
-  Should be an inline option to pick a browser to import settings from via drop down
-  "Import" button is disabled unless an valid option is chosen
-  Import button should lead to next step of on-boarding process

- https://brave-ui-bpu9szh0x.now.sh/?path=/story/feature-components-welcome--page

## Integration
  

imptrx:
Should be no breaking changes in brave-core as it's mostly storybook changes plus duplication of a component 🙈

bsclifton:
This is amazing  😄 Doesn't look like it would be too hard to integrate into https://github.com/brave/brave-core/pull/2679 after this is done

PR #510, opened 2019-07-05
add multiple views for shields

cezaraugusto:
see https://brave-ui-4200l5i5w.now.sh/?path=/story/feature-components-shields--panel
address https://github.com/brave/brave-browser/issues/1196

### Test Plan 
See https://github.com/brave/brave-browser/issues/1196 for design spec

1. Initial state for Shields should be the "simple view" (no option for controlling resources blocked)
2. First time switching to "advanced view" should show a warning. This warning should show only. one time, i.e. toggling views again should not show it.
3. Should be able to toggle simple/advanced views via link in Shields footer.


src/components/icons/alert-message-o.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+import * as React from 'react'
+import GenerateComponentForGraphic from './template'
+
+// Direct access to SVG
+export const Graphic = (
+  <svg width='32' height='32' xmlns='http://www.w3.org/2000/svg' xmlnsXlink='http://www.w3.org/1999/xlink'>
+    <defs>
+      <path d='M15.29 17.71c-.18-.19-.29-.45-.29-.71 0-.13.03-.26.08-.38.05-.13.12-.23.21-.33.09-.09.2-.16.33-.21.24-.1.52-.11.76 0 .13.05.24.12.33.21.09.09.16.2.21.33.05.12.08.25.08.38 0 .26-.11.52-.29.71-.19.18-.45.29-.71.29-.13 0-.26-.03-.38-.08a.933.933 0 0 1-.33-.21zM15 14v-4a1 1 0 0 1 2 0v4a1 1 0 0 1-2 0zm11 6H13c-.019 0-.035.01-.053.011-.071.004-.136.022-.205.041-.059.016-.117.028-.171.054-.055.026-.101.065-.152.102-.056.041-.111.08-.158.133-.013.014-.03.02-.042.035L10 23.149V21a1 1 0 0 0-1-1H6V7h20v13zm0-15H6c-1.103 0-2 .897-2 2v13c0 1.103.897 2 2 2h2v4a1 1 0 0 0 1.781.625L13.48 22H26c1.103 0 2-.897 2-2V7c0-1.103-.897-2-2-2z'
+      id='a' />
```
petemill: Please convert this to not use defs with an explicit ID (especially with #a which will almost certainly exist on the page already if there are 2 such graphics). This looks like it can be converted to a `<path` directly in the SVG, with the `fill` and `fillRule` from the `<use /` made inline to the `<path`.

cezaraugusto:
Thanks for checking @petemill. I tagged the usual reviewers for Shields which is why I didn't add you at first, but happy you checked.

re  https://github.com/brave/brave-ui/pull/510#pullrequestreview-260284704

1) Didn't try it specifically but seems doable. extension popup seems to handle height changes well.  
2) I did the first version of our dialog using the native `<dialog>` element. Got some issues with TypeScript back then and refactored to be a bunch of `<div>`s. I agree with you we should use it and  reconsider making a shared component, but I don't think I can come up with a good version in this PR.

src/components/icons/alert-message-o.tsx
cezaraugusto: good find, updated

PR #487, opened 2019-05-31
Welcome dark theme

rossmoody:
Addresses: https://github.com/brave/brave-browser/issues/4657
Addresses: https://github.com/brave/brave-browser/issues/3911

Now: https://brave-ui-en211eort.now.sh

## Before

image

## After

image

image

PR #512, opened 2019-07-09
✨ Inline theme selection at on-boarding

imptrx:
Spec: https://github.com/brave/brave-browser/issues/4992
Quick View:
welcome-theme

## Changes
- inline option to select theme during onboarding

## Test plan
- go to brave://welcome
- at the 4th screen during the on-boarding process you now have a dropdown to select theme
- button is disabled unless a valid option is selected
- clicking the confirm should go to next screen

- https://brave-ui-iwtuqykho.now.sh/?path=/story/feature-components-welcome--page

## Integration
  
### DIFF
--- a/stories/features/welcome/page/fakeLocale.ts
+++ b/stories/features/welcome/page/fakeLocale.ts
@@ -25,7 +25,12 @@ const locale = {
   fakeSearchProvider2: 'Google (default)',
   // fourth screen
   chooseYourTheme: 'Choose your color theme',
+  selectTheme: 'Select a theme',
   findToolbarTheme: 'Set the color of your toolbar by selecting the light or dark option from the settings panel.',
+  themeOption1: 'Light',
+  themeOption2: 'Dark',
+  themeOption3: 'System theme (default)',
+
   // fifth screen
   protectYourPrivacy: 'Manage your shields',
   adjustProtectionLevel: 'Protect against privacy-invading ads and trackers while browsing with Brave Shields. Set Shields to "down" if a site doesn’t seem to be working properly.',
--- a/stories/features/welcome/page/screens/themeBox.tsx
+++ b/stories/features/welcome/page/screens/themeBox.tsx
@@ -5,7 +5,8 @@
 import * as React from 'react'
 
 // Feature-specific components
-import { Content, Title, Paragraph, PrimaryButton } from '../../../../../src/features/welcome/'
+import { Content, Title, Paragraph, PrimaryButton, SelectGrid } from '../../../../../src/features/welcome/'
+import { SelectBox } from '../../../../../src/features/welcome'
 
 // Utils
 import locale from '../fakeLocale'
@@ -19,9 +20,30 @@ interface Props {
   onClick: () => void
 }
 
-export default class ThemingBox extends React.PureComponent<Props, {}> {
+interface State {
+  themeSelected: boolean
+}
+
+export default class ThemingBox extends React.PureComponent<Props, State> {
+
+  constructor (props: Props) {
+    super(props)
+    this.state = {
+      themeSelected: false
+    }
+  }
+
+  onClickSelectTheme = () => {
+    this.props.onClick()
+  }
+
+  onChangeThemeOption = (event: React.ChangeEvent<HTMLSelectElement>) => {
+    this.setState({ themeSelected: event.target.value !== '' })
+  }
+
   render () {
-    const { index, currentScreen, onClick } = this.props
+    const { index, currentScreen } = this.props
+    const { themeSelected } = this.state
     return (
       <Content
         zIndex={index}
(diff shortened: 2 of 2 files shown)


stories/features/welcome/page/fakeLocale.ts
```diff
@@ -25,7 +25,12 @@ const locale = {
   fakeSearchProvider2: 'Google (default)',
   // fourth screen
   chooseYourTheme: 'Choose your color theme',
+  selectTheme: 'Select a theme',
   findToolbarTheme: 'Set the color of your toolbar by selecting the light or dark option from the settings panel.',
+  themeOption1: 'Dark',
+  themeOption2: 'Light',
+  themeOption3: 'Same as macOS',
```
rossmoody: i know this is fake but let's do:
- Light
- Dark
- System

stories/features/welcome/page/screens/themeBox.tsx
```diff
@@ -32,13 +54,22 @@ export default class ThemingBox extends React.PureComponent<Props, {}> {
...
-            size='large'
-            text={locale.theme}
-            onClick={onClick}
-          />
+        <SelectGrid>
+            <SelectBox onChange={this.onChangeThemeOption}>
+              <option value=''>{locale.selectTheme}</option>
+              <option value='Dark'>{locale.themeOption1}</option>
+              <option value='Light'>{locale.themeOption2}</option>
+              <option value='System'>{locale.themeOption3}</option>
```
rossmoody: let's do:
- Light
- Dark
- System

stories/features/welcome/page/fakeLocale.ts
rossmoody: `System`

imptrx:
Will try to get this and #487 in at the same time to maximize the difference when the user changes the settings 😃

stories/features/welcome/page/screens/themeBox.tsx
petemill: How about we bring in the strings from the product, and create an issue to change the existing ones to what you suggest @rossmoody ? Then it's consistent with Settings page either way?

petemill:
One concern I have with this UX @rossmoody is that the user may feel that they *have to* pick a theme, and not know that 'Same as {macOS,Windows}' is the default. Therefore they may explicitly choose Light / Dark, and not know that they could have it follow the OS when they changed it.

To be honest, I think having this choice on the welcome screen had high value, before we matched system theme, especially if it wasn't a drop-down but was a visual choice, but now it seems confusing perhaps.

I don't know if there's a way we can offer this tri-state in a way that doesn't prompt users to override the default unneccessarily?

petemill:
Also, how about we don't have a `Confirm` button, but have the change previewed and set immediately?

rossmoody:
The original goal of this screen was just to show users that they could take advantage of dark mode cause it wasn't offered in Muon. Since we're defaulting to following system setting, the delight and surprise of this screen will more likely be discovered in screen number one and I don't think (though I wish we could test) that many people opt to use light them in dark system settings. I do think (for the moment) this screen is worth keeping though because Chrome doesn't allow light system theme and dark browser theme and we do. (That said I also think this could get the boot in the next rev of the welcome flow for something stronger).

Maybe this could be solved immediately with a default designation in the list item: `System Theme (default)`

I much prefer your idea though. Maybe there are the three options but `Same as System Theme (default)` is preselected when the user arrives. Maybe we do the updates as a followup?

rossmoody:
I started experimenting with the new approach here but it feels like it's own initiative that needs review and probably shouldn't block this PR. 

image

I'd suggest we move forward with this for now and add a `(default)` to the System list item so it will read:

`System theme (default)` @imptrx @rebron

rebron:
@rossmoody Sounds good. Let's go with System theme (default)

petemill:
Should we keep or remove the Confirm button? Surely without it we won't see the browser theme change from light to dark in realtime?

rossmoody:
Keep for now, it's really just to give the user a place to feel like they solidified the decision though I'm fairly certain the dropdown will behave like it does in settings. If Dark theme is selected in the dropdown the theme will change to dark and then "Confirming" it will advance to the next screen.

petemill:
> Keep for now, it's really just to give the user a place to feel like they solidified the decision though I'm fairly certain the dropdown will behave like it does in settings. If Dark theme is selected in the dropdown the theme will change to dark and then "Confirming" it will advance to the next screen.

So just to _confirm_ that "Confirm" won't actually save the setting

rossmoody:
Right, under the hood it's set when the selection is made.

imptrx:
These changes will need to be pulled into https://github.com/brave/brave-core/pull/2909 since the migration of welcome specific ui: brave/brave-core#2896

PR #515, opened 2019-07-17
Fixing detail view and settings page logic error (boxMobile) [Android]

ryanml:
Related: https://github.com/brave/browser-android-tabs/issues/1858

Going back from `settings (ex: rewards/ads-settings) -> detailView (ex: rewards/ads)` was not working as expected when pushing the Android back button. Entries in the history state are fine, there was just a boolean logic problem when doing so.

### DIFF
--- a/src/features/rewards/mobile/boxMobile/index.tsx
+++ b/src/features/rewards/mobile/boxMobile/index.tsx
@@ -98,13 +98,36 @@ export default class BoxMobile extends React.PureComponent<Props, State> {
     }
   }
 
+  /*
+   * Sets state given the view that should be shown.
+   *
+   * When user is on the index page: (brave://rewards)
+   * (state) { detailView: false, settings: false }
+   *
+   * When user is on a detail view: (brave://rewards/ads)
+   * (state) { detailView: true, settings: false }
+   *
+   * When user is on a settings view: (brave://rewards/ads-settings)
+   * (state) { detailView: true, settings: true }
+   */
   setView = (view: string, updateHistory: boolean = true) => {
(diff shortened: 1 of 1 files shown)


src/features/rewards/mobile/boxMobile/index.tsx
```diff
@@ -99,12 +99,8 @@ export default class BoxMobile extends React.PureComponent<Props, State> {
   }
 
   setView = (view: string, updateHistory: boolean = true) => {
+    const isDetailView = true
```
ryanml: Via this method, `detailView` should always be true. The logic for handling the index `/` view happens elsewhere

The truth table for this should be:

Index page `rewards/`:
`{ detailView: false, settings: false }`
Detail page `rewards/ads`:
`{ detailView: true, settings: false }`
Detail page `rewards/ads-settings`:
`{ detailView: true, settings: true }`

petemill: @ryanml do you feel this kind of explanation should go in as a code comment? Aim is for everything to be understandable when reading through, and I'm not sure it is without this context.

ryanml: @petemill can update with code commentary, I also pushed another adjustment:

The previous logic would attempt to (incorrectly) toggle the states (so if you were on settings and wanted to go back `setView('settings')` was called to `setState({ settings: !settings})` this didn't work in every case.

I've updated `setView` to explicitly take the view that is to be shown. So in the above case, going back from the settings page would require `setView('detailView')`, and going back from the detail view will require `setView('index')`

PR #541, opened 2019-09-18
Use new button styles

### LINKED ISSUE #536: Restyle buttons according to updated mockups
The current button designs are outdated.

They should be updated to look like the ones in this mockup:

Button Specs

In dark mode, "Default" style buttons should be styled as follows: white text and #c2c4cf border, Hover style should be #FF7654 bold, weight: 700 text and #FF7654 border. Focused treatment is white text, #FF7654 border with 3px #FF7654 40% alpha border. and pressed state is #FF7654 text and #FF7654 border with 3px #FF7654 40% alpha border

"Primary" style buttons are identical between dark and light themes.

antonok-edm:
## Changes

Fixes #536.

Addresses concerns from #486 as well.

With this PR, buttons are all themed using a common `Button` component, and have a visual update as described in #536.

This will require changes in https://github.com/brave/brave-core wherever buttons are used as well.

## Test plan

Storybook stories and snapshots have been updated to use the new buttons. I've done a scan through all of the usages of buttons to verify that everything looks acceptable, but another pair of eyes would be appreciated.

- Primarily http://localhost:9091/?path=/story/components--button, but most of the stories include some kind of button and have also been updated accordingly.

## Integration
- [x] Does this contain changes to src/components or src/
  - [x] Incompatible API change to something existing _(major version increase)_
  
  - [x] Are there non backwards-compatible changes required for brave-core? **Do not merge until brave-core PR is approvable.** Link to brave-core PR:

### DIFF
--- a/src/components/buttonsIndicators/button/index.tsx
+++ b/src/components/buttonsIndicators/button/index.tsx
@@ -4,6 +4,7 @@
 
 import * as React from 'react'
 import {
+  DefaultButton,
   PrimaryButton,
   SecondaryButton,
   TertiaryButton,
@@ -14,10 +15,8 @@ import {
 export interface Props {
   text: string
   size?: Size
-  type?: Type
   brand?: Brand
   level?: Level
-  main?: boolean
   onClick?: () => void
   id?: string
   disabled?: boolean
@@ -25,8 +24,7 @@ export interface Props {
   className?: string
 }
 
-export type Level = 'primary' | 'secondary' | 'tertiary'
-export type Type = 'default' | 'accent' | 'warn' | 'subtle'
+export type Level = 'default' | 'primary' | 'secondary' | 'tertiary'
 export type Brand = 'brave' | 'rewards'
 export type Size = 'call-to-action' | 'large' | 'medium' | 'small'
 
@@ -36,13 +34,13 @@ export default class ThemedButton extends React.PureComponent<Props, {}> {
   static defaultProps = {
     brand: 'brave',
     size: 'medium',
-    type: 'default',
-    level: 'primary',
-    main: false
+    level: 'default'
   }
 
   getButtonComponent () {
     switch (this.props.level) {
+      case 'default':
+        return DefaultButton
       case 'primary':
         return PrimaryButton
       case 'secondary':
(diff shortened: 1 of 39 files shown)


src/components/buttonsIndicators/button/index.tsx
```diff
@@ -14,19 +15,16 @@ import {
...
   onClick?: () => void
   id?: string
   disabled?: boolean
   icon?: {image: React.ReactNode, position: 'before' | 'after'}
   className?: string
 }
 
-export type Level = 'primary' | 'secondary' | 'tertiary'
-export type Type = 'default' | 'accent' | 'warn' | 'subtle'
+export type Level = 'default' | 'primary' | 'secondary' | 'tertiary'
```
petemill: I'm not sure it's clear to me that we should have a `default` value here - it's not understandable how this semantically fits in between the hierarchy of primary / secondary / tertiary

petemill:
Please add a storybook link via running `now -p` on your local branch.

src/components/buttonsIndicators/button/index.tsx
antonok-edm: I just took that directly from abstract, cc @karenkliu

petemill: @karenkliu @antonok-edm we should probably sync-up as it seems that whilst this has removed some variations, it seems more complicated to use. Right now we have properties that easily allow a developer or designer to know when to use which button:

- 'level' which indicates the importance of the button (most important in the UI vs one of many equal actions) **and affects the style of the button**,

image


- 'type' which indicates the type of action (positive, negative, primary brand action) **and just really affects the color**

image

These two properties are independent and could allow for all combinations.

In this new version we have a single property called 'Levels' in which I would expect the values to be hierarchical and has 4 variations - `primary`, `secondary`, `tertiary` and `default`. I wouldn't know when to use 'default' vs the hierarchical values. I would expect most buttons which aren't Primary actions to be Secondary actions, but now I'm guessing they should be Default actions and it's unclear when to use Secondary vs Tertiary.

image

IMO we shouldn't mix metaphors here - either keep the semantic definitions (and ideally backwards-compatible) or choose new ones.


As another aside, these new button styles are slightly different to the other new button styles that Ross implemented in brave-core's WebUI at https://github.com/brave/brave-core/pull/2778. Those ones have a slight hover effect on 'pressed' and also a nicer focus outline style (that last one we should definitely adopt).

PR #544, opened 2019-10-04
allow any kind of theme for themeProvider and helper methods

cezaraugusto:
This allows `ThemedStyledComponentsModule` to receive any theme as a prop. Needed for https://github.com/brave/brave-core/pull/3300.

The reasoning I had is that we can type feature-specific themes in there instead of forcing a theme in root level. I'm open for discussion regarding better options.

### DIFF
--- a/src/theme/index.tsx
+++ b/src/theme/index.tsx
@@ -1,7 +1,7 @@
 // @ts-ignore: Needed for ThemeProvider
 import * as React from 'react'
 import * as styledComponents from 'styled-components'
-import IThemeProps, { BraveThemedStyledProps as ThemedStyledProps } from './theme-interface'
+import { BraveThemedStyledProps as ThemedStyledProps } from './theme-interface'
 // theme for testing
 import TestTheme from './brave-default'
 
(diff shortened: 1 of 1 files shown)


src/theme/index.tsx
```diff
@@ -17,7 +17,7 @@ const {
   //   see: https://github.com/palantir/tslint/issues/3505
   //   It's possibly due to the rule upstream in tslint-config-standard
   // tslint:disable-next-line
-} = styledComponents as styledComponents.ThemedStyledComponentsModule<IThemeProps>
+} = styledComponents as styledComponents.ThemedStyledComponentsModule<any>
```
petemill: There's got to be a better way than just using `any` here, right? Wouldn't this lose all intellisense for theme props in all brave components (`${p => p.theme.xyz}`)? Perhaps `Partial<IThemeProps>` would work?

petemill: Partial won't work. `<IThemeProps & any>` might, but not sure what that will get us. Maybe it's good. Either way seem comment on https://github.com/brave/brave-core/pull/3300 as to whether this is needed yet.

cezaraugusto:
follow-up issue https://github.com/brave/brave-ui/issues/545

PR #555, opened 2020-01-16
Remove Shields from brave-ui

cezaraugusto:
> Note: Shields theme left un-removed until we define a good place for it in brave-core.

_ciao_ Shields :)


cezaraugusto:
I am hitting the merge button since this change doesn't affect existing functionality and is only removing dead code. I did a manual test in brave-core, and it builds and runs as expected. Also, CI is green.

PR #558, opened 2020-03-23
Add icons for "credit card" and "add"

zenparsing:
## Changes

Adds credit card and "add" (plus) icons for use in SKU work.

- 

## Integration
  

src/components/icons/index.tsx
```diff
@@ -17,6 +17,7 @@ function RotatedIconComponent (
 }
 
 // simple
+export { default as Add } from './add'
```
NejcZdovc: Maybe this should be plus instead of add, also it needs Icon at the end

zenparsing: Fixed

PR #589, opened 2021-06-08
adds v2 icons to story and generator

AlanBreck:
This is a temporary fix in order to add v2 icons (currently only used in search) to the icons story book.

NOTE: that the icon sprite aspect assumes the source folder contains ALL icons (not just updated ones) and overwrites the existing `brave-icons-svg.js` file. This ONLY impacts `v2` icons.


andikaleonardo (approved):
gg

andikaleonardo (approved):

src/components/icons_v2/Done.tsx
```diff
@@ -0,0 +1,19 @@
...
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License. v. 2.0. If a copy of the MPL was not distributed with this file.
+* You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import * as React from 'react'
+import GenerateComponentForGraphic from '../icon-template'
+
+// Direct access to SVG
+export const Graphic = (
+<svg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 25 24'>
```
petemill, line 12: hmm, so most (all?) of these viewBoxes are 25x24, but we're rendering them in a 32x32 svg. Actually what are we doing with that icon template? I just noticed it's 2 layers of svg, seems like it's rendering
```
<svg viewBox='0 0 32 32'>
  <svg viewBox='0 0 25 24'>
  ...
```
I think that may cause some rendering issues? I feel like we should either have the template function accept the viewbox param, and optionally have the JSX for the graphic just be the children, e.g.
```
<>
  <path fill-rule='evenodd' d='m11.6817 14.3813-2.30039-1.9697c-.38322-.3281-.95986-.2834-1.28798.0998-.32812.3832-.28345.9599.09976 1.288l2.68131 2.2958c.6131.5249 1.5449.4088 2.0105-.2506l4.8215-6.82895c.291-.41212.1928-.98209-.2194-1.27307-.4121-.29098-.9821-.19277-1.273.21936l-4.5323 6.41936z'
    clip-rule='evenodd' />
</>
```

Have you noticed any issue in the storybook or other react with these new icons @AlanBreck ?

petemill (commented):
Nice work! I think we have an issue with the viewBox being different than the template, but happy to merge and fix iteratively if there isn't an easy solution.

src/components/icons_v2/Done.tsx
AlanBreck: The viewbox simply manages the internal reference point for the coordinates, right, so it really wouldn't matter what actual dimensions we use? That's the behavior I've observed, at least.

andikaleonardo (approved):

src/components/icons_v2/Done.tsx
AlanBreck: Strangely, I'm not seeing any rendering issues. Additionally, when inspecting the icons in Storybook, they only show one layer of SVG.

If possible, I'd prefer to merge and fix iteratively if we encounter issues.

AlanBreck:
Just pinging on the status of this PR. Thanks!

PR #607, opened 2022-06-24
Remove muli

petemill:
## Changes
Removes all occurances of the Muli font and replaces with Poppins. The biggest change is in the theme file, which other projects use, but really mainly (only?) brave-core.

## Test plan
Make sure storybook runs with no error. I verified this.
Main visual (manual) tests will be in brave-core.

- https://brave-ui-petemill.vercel.app/

## Integration
  

boocmp:
I don't actually understand what is StoryBook and how I can use it. The last time I showed an build from my PR to @aguscruiz, he recommended to change the font weight for Poppins. I'm not sure, but it looks like we need to change the weights in the storybook.

petemill:
@boocmp I think we'll need to build brave-core with this brave-ui ref in order to see what changes would be neccessary, and I think most or all of those changes would be in brave-core code. I'll push to your branch in https://github.com/brave/brave-core/pull/12933 an update to the brave-ui ref so that we can get a browser build and Storybook

boocmp (approved):

PR #629, opened 2023-01-26
Move all theme variables to a parent property named "legacy"

petemill:
…so that any uses of this theme are seen as deprecated and do not conflict with modern theme properties

### DIFF
--- a/definitions/styled-components-theme.d.ts
+++ b/definitions/styled-components-theme.d.ts
@@ -1,0 +1,13 @@
+// Copyright (c) 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import IBraveUITheme from '../src/theme/theme-interface'
+
+declare module 'styled-components' {
+  export interface DefaultTheme {
+    // Marked legacy as this repo is deprecated
+    legacy: IBraveUITheme
+  }
+}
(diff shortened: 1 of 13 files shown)


definitions/styled-components-theme.d.ts
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) 2023 The Brave Authors. All rights reserved.
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this file,
+// You can obtain one at https://mozilla.org/MPL/2.0/.
+
+import IBraveUITheme from '../src/theme/theme-interface'
```
fallaciousreasoning, line 6: This import doesn't look safe to me - what if the library is used something that isn't `brave-core`? Presumably this was already getting the `IBraveUITheme` interface from somewhere though - I guess this is fine if it's what we were doing already, just wanted to double check.

fallaciousreasoning (approved):
LGTM, assuming the one comment is okay.

definitions/styled-components-theme.d.ts
petemill: This was really just for the internal storybook. A consuming app which doesn't define their own theme should not end up using this custom definition for a transient dependency module, right? Or would it?

AlanBreck (approved):

definitions/styled-components-theme.d.ts
petemill: I don't think we (should) see any more contributions to this repo (at least, any that aren't bug fixes). As for brave-core compat, I have renamed all uses there to `.theme.legacy.`

petemill:
This is not urgently needed since the new design system tokens don't get put on styled-component's `theme` prop. They are a static map of key to css variable name string.

PR #642, opened 2023-08-04
Setup for @brave/brave-ui publish

petemill:
Brave does not own npm's brave-ui package.

This package is deprecated, but we need 1 more publish to make sure dependents are on a JS-compiled version and don't need to compile typescript in node_modules, which is causing some issues.

Removes storybook dependencies are they are too out of date for npm to work without package-lock format v1 (npm is now on v3). NPM refuses to install due to bad peer dependency matches.


petemill:
Updated to make sure JS files have commonjs module format

package.json
```diff
@@ -30,13 +31,10 @@
     "lint": "npm run lint-css && npm run lint-ts",
     "lint-css": "stylelint './src/**/*.{ts,tsx}'",
     "lint-ts": "tslint --project tsconfig-storybook.json 'src/**/*.{ts,tsx}' 'stories/**/*.{ts,tsx}'",
-    "copy": "ncp src package --filter='^(?!(.*)(__snapshots__|spec.tsx)$).*$' && cpr package.json package/ && cpr README.md package/",
-    "build": "rimraf package && npm run copy && lerna exec -- tsc",
+    "copy": "cpr package.json package/ && cpr README.md package/ && cpr LICENSE.md package/",
+    "build": "rimraf package && npm run copy && tsc --outDir package",
     "publish": "npm run build && lerna publish",
     "now-build": "npm run storybook-build",
```
fallaciousreasoning, line 37: This can probably be removed too

package.json
```diff
@@ -109,22 +91,12 @@
...
-      "lint"
-    ]
-  },
-  "pre-push": {
-    "run": [
-      "lint",
-      "test-unit-all"
-    ]
+    "react": "^16.0.0",
+    "styled-components": "^5.2.1"
```
fallaciousreasoning, line 99: :+1:

package.json
```diff
@@ -109,22 +91,12 @@
     "tslint": "^6.1.3",
     "tslint-config-standard": "^9.0.0",
     "tslint-react": "^5.0.0",
-    "typescript": "^3.0.1",
-    "typescript-plugin-styled-components": "0.0.6"
+    "typescript": "^4.9.5",
```
fallaciousreasoning, line 94: We're at 5.1 now, so 3.0 was a loooong time ago :D

src/old/v1/image/index.tsx
```diff
@@ -15,14 +15,12 @@ export interface ImageTheme {
   padding?: string
 }
 
-export interface ImageProps {
+export type ImageProps = React.PropsWithChildren<{
   customStyle?: ImageTheme
   id?: string
-  children?: HTMLSourceElement
```
fallaciousreasoning, line 21: I think `children?: HTMLSourceElement | HTMLSourceElement[]` would be better than `React.PropsWithChildren` as they're the only children the picture element can take, right?

fallaciousreasoning: Oh, just noticed this is used for the `figure` too. Ah well :laughing:

fallaciousreasoning (approved):
LGTM

petemill:
Reverted back to esmodule as things like `export * from ...` was causing errors in the consuming project, and I fixed jest in the consumer

fallaciousreasoning (approved):
Still LGTM

AlanBreck (approved):

